PR #60, opened 2015-03-22
A few updates, new libs and options

wiiaboo:


jb-alvarado:
Hi Ricardo,
wow that is a big list :). Thank you for that. Only some little points:
- When I remember me right, the ligmagic was also used by sox  
- I am not sure that people are happy when we put the faac out from ffmpeg. For me is good, but some like to use different aac encoders in ffmpeg.
- is i still possible to build the static exes from x264 and x265?

I also have made a update yesterday - is is no problem for the patches?

wiiaboo:
>
>
>    - When I remember me right, the ligmagic was also used by sox
>
> Ok, removed libmagic from that commit. Only wxWidgets is optional.

>
>    - I am not sure that people are happy when we put the faac out from
>    ffmpeg. For me is good, but some like to use different aac encoders in
>    ffmpeg.
>
> It's really really an awful encoder. Even ffmpeg's native encoder is
better. And if you're compiling a non-free encoder, fdk-aac is much better.
About as good as Apple's AAC.

>
>    - is i still possible to build the static exes from x264 and x265?
>
> Yes, that's why I separated the options. Either way x264 and x265 are
always built, it's only a matter if a binary is built or just the libraries.
Maybe the options should be "binary; libs; not compile at all"?

jb-alvarado:
Ok I am convinced with faac :). Then I do now the merge and run a test, hopefully everything works after :D.

PR #62, opened 2015-03-23
Change most versioned libs to use pkg-config instead of just checking if lib exists

wiiaboo:
Missing libs:
- openjpeg
- libgnurx
- libmad

First compilation from master but never get updates:
- xavs (doesn't get many updates anyway)
- mediainfo
- decklink

The rest all get compiled from vcs so they always get updated when needed.

### DIFF
--- a/media-suite_compile.sh
+++ b/media-suite_compile.sh
@@ -813,30 +813,24 @@ fi
 
 cd $LOCALBUILDDIR
 
-if [ -f "$LOCALDESTDIR/bin-audio/fdkaac.exe" ]; then
+do_git "https://github.com/nu774/fdkaac" bin-fdk-aac
+
+if [[ $compile == "true" ]]; then
+    if [[ ! -f ./configure ]]; then
+        autoreconf -i
+    fi
+
+    ./configure --build=$targetBuild --host=$targetHost --prefix=$LOCALDESTDIR --bindir=$LOCALDESTDIR/bin-audio
+
+    make -j $cpuCount
+    make install
+
+    do_checkIfExist bin-fdk-aac bin-audio/fdkaac.exe
+    compile="false"
+else
     echo -------------------------------------------------
-    echo "bin-fdk-aac is already compiled"
+    echo "bin-fdk-aac is already up to date"
     echo -------------------------------------------------
-    else
-        echo -ne "\033]0;compile fdk-aac $bits\007"
-        rm -rf bin-fdk-aac
-
-        wget --tries=20 --retry-connrefused --waitretry=2 --no-check-certificate -c https://github.com/nu774/fdkaac/archive/master.zip -O bin-fdk-aac.zip
-        unzip bin-fdk-aac.zip
-        rm bin-fdk-aac.zip
-        mv fdkaac-master bin-fdk-aac
-        cd bin-fdk-aac
-
-        if [[ ! -f ./configure ]]; then
-            autoreconf -i
-        fi
-
-        ./configure --build=$targetBuild --host=$targetHost --prefix=$LOCALDESTDIR --bindir=$LOCALDESTDIR/bin-audio
-
-        make -j $cpuCount
-        make install
-
-        do_checkIfExist bin-fdk-aac bin-audio/fdkaac.exe
 fi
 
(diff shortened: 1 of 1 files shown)


wiiaboo:
MediaInfo is also present in Github, but ZenLib, MediaInfoLib and MediaInfo are all separated repos.

Maybe we could keep MediaInfo updated by checking the version too, if using the git repos is too unstable.

jb-alvarado:
Thanks again, the changes is added. I think it is enough to check the mediainfo version. We don't need a git version from that. It was a nice idea to check this. Sometimes it makes problems, when old header files from a build is in the system. That is way I have this delete options by ffmpeg and some other builds. So we need to have a look on that. decklink is not so easy to update, because blackmagic only allow me to have this specific files in my script, not a download link to the sdk. And for now I need to generate the include file from decklink, so bedder we let this like it is.

wiiaboo:
```
if [[ `mediainfo --version | grep "MediaInfoLib - v0.7.72"` ]]
```
instead of `media-suite_compile.sh` line 1333 should do it then.

jb-alvarado:
Ah now I remember... This doesn't work, because every time media player is build new it get the lasted source code. it is dynamic, so we can not check the version number. When the we need to check the version from Internet.

wiiaboo:
Which media player are you talking about?

jb-alvarado:
Sorry I mean mediainfo... A other question, ffmpeg will not build by me anymore:

libavcodec/libavcodec.a(qpeldsp_init.o):qpeldsp_init.c:(.text+0x260d): undefined reference to `ff_avg_mpeg4_qpel8_v_lowpass_mmxext'

Any idea?

wiiaboo:
I didn't compile ffmbc so I assumed it was just like ffmpeg.
I'm compiling everything from scratch now to see what's wrong.

jb-alvarado:
The script runs by me and now I think it works everything. ffmpeg 32 bit I need to try again. And the check from ffmbc I change to the old way. Maybe you have here a better way. I also fixed some little things - copy & paste errors :).

wiiaboo:
To check ffmbc version you need to redirect stderr to stdout, so like this should work:
```
if [[ `ffmbc.exe 2>&1 | grep "version 0.7.3"` ]]; then
```
like a52dec.

wiiaboo:
Btw, apparently flac detecting doesn't work because pacman installs flac somewhere.

jb-alvarado:
I think we can use: 
´pkg-config --define-variable=PKG_CONFIG_PATH="$LOCALDESTDIR/lib/pkgconfig" --modversion flac´

wiiaboo:
If mingw's flac is not enough, sure. Why do we compile flac? Problems with shared lib?

jb-alvarado:
Somebody wonts the static exe, that is why I have them in.

wiiaboo:
Makes sense, I use it too.
Since we're compiling libsndfile now we can probably remove that from pac-mingw32/64-new.pak

jb-alvarado:
Ah now I'm remember, I had change it to pacman since some time. Bad you say you fix it, was it not correct before? Then yes we can delete it from pacman.

wiiaboo:
I added sndfile again because of libbs2b not working with mingw's for some reason.

wiiaboo:
For some reason ffmpeg compiling keeps breaking on avisynth for me. Dunno what caused it.

wiiaboo:
This commit: https://github.com/FFmpeg/FFmpeg/commit/a8c99205ca8703bd849efae13fcf844315c7147d messes with avisynth building.

jb-alvarado:
Am 25.03.2015 um 00:19 schrieb Ricardo Constantino:
Yes, I had write it yesterday to the back tracker.
>
> This commit: FFmpeg/FFmpeg@a8c9920 
> <https://github.com/FFmpeg/FFmpeg/commit/a8c99205ca8703bd849efae13fcf844315c7147d> 
> messes with avisynth building.
>
-- 
_________________________________________________

Web: http://www.pixelcrusher.de
Blog: http://blog.pixelcrusher.de

Skype: jb_alvarado
Fax/UMS: +4932224128180

wiiaboo:
Everything seems to be working fine now.

PR #65, opened 2015-03-27
Git improve

wiiaboo:
Haven't tested yet. Will do it tomorrow.
Request is just so we can discuss if there's something you don't want.


wiiaboo:
Potential optional components:
[X] openjpeg
[ ] rtmpdump
[ ] libilbc
[ ] vo-aacenc
[ ] opencore amr
[ ] libgme
[ ] twolame
[ ] libbs2b
[ ] vpx (build binaries or just libs)
[ ] utvideo
[ ] xavs
[ ] mediainfo
[ ] caca
[ ] zvbi
[ ] decklink

Picked these libs because many are encoders only and maybe most people don't need them. I picked the ones I personally don't need but maybe we can narrow them so people don't have to make 20+ choices when first opening the .bat?

jb-alvarado:
Hi Ricardo,
sorry for my late response, at the moment is not so easy for me to found free time and maybe this get not better in the future. I like the idea to have ffmpeg more customized, but I think it is not so nice when people have to select many options. Maybe we can make a option where we ask that somebody want to chose different libs for ffmpeg and when you select yes then comes more options. When you select no, the all libs gets build. You know what I mean?

mediainfo we can make optional that is a good idea.
Using only one build dir is also good, but then for everything, also for the normal sources. But we need to see how we can realize that. All other options are ok for me.

wiiaboo:
> Maybe we can make a option where we ask that somebody want to chose different libs for ffmpeg and when you select yes then comes more options. When you select no, the all libs gets build.

That's a good idea.

I'll save that until I've made at least the above optional. Up until now, no problems compiling, but i'm only testing static ffmpeg. But I'm not really changing anything, just making things optional.

>Using only one build dir is also good, but then for everything, also for the normal sources.

For that to work we'd need some kind of variable or textfile that can mark whether something has been updated or not for the non-versioned stuff (hg/git).

jb-alvarado:
vpx has also the option to build 12bit encoder, maybe we can put this also in.

wiiaboo:
Heh, I don't think there's that much benefit above 10-bit, but since we're adding this much choice, why not

jb-alvarado:
Is there a way to merge specific commits?

wiiaboo:
I moved all the optional-libs components to another specific branch. Try this new version instead.

jb-alvarado:
Thank you Ricardo!

PR #66, opened 2015-04-08
Compile luajit-git and remove lua

wiiaboo:
Fixes mpv compilation with lua which was broken since update to Lua 5.3.
MinGW's luajit-git doesn't work.


jb-alvarado:
Thank you, Ricardo!

PR #72, opened 2015-04-11
Use only one build dir

wiiaboo:
Only tested this with my usecase (64-bit only).
Will test with 32-bit as well tomorrow.

The separate "build_successful$bits" mean that in case something happens during 64-bit compilation the 32-bit won't get compiled again.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -956,7 +956,7 @@ if exist %instdir%\%msys2%\mingw64\lib\xvidcore.dll.a (
 Setlocal DisableDelayedExpansion
 
 if %build32%==yes (
-    if not exist %instdir%\build32 mkdir %instdir%\build32
+    if not exist %instdir%\build mkdir %instdir%\build
     if not exist %instdir%\local32\share (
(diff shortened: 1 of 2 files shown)


jb-alvarado:
look very good! I will run a test tomorrow.

jb-alvarado:
have you a idea how we can use this for the normal zipped sources to?

wiiaboo:
Don't delete them with do_checkIfExist, delete them at the end of the
script and then change every zip source to use the existing dir instead of
downloading.

jb-alvarado:
And maybe we need to check to, witch repos have distclean support instead of clean.

jb-alvarado:
I think the "touch build_successful$bits" don't work right at the moment. By me it has run two times the same target.

wiiaboo:
Yeah, I think it's because you need deletesource to be on. Just replace the
"else" before and switch to "fi"

wiiaboo:
What do you think of https://github.com/jb-alvarado/media-autobuild_suite/commit/fa1d1f63f4ede5f185e516f6168f90291a2e7395 ?

PR #73, opened 2015-04-12
Fix/Make do_checkIfExist more efficient

wiiaboo:


jb-alvarado:
When we delete the source folders at the end, then we can put a check in the do_wget function to, that it only download the source new when the folder not exist.

PR #76, opened 2015-04-12
Wget improve

wiiaboo:
You can delete the other. This one has same and more changes and it's in a separate branch.

About using the new do_wget with zip and 7z, I don't think it's worth it. You need more lines of code to support the couple of files that are not .tar.* than to keep using the old do_wget.

In the future a better and more efficient way can be used, but for now, this is good enough, IMO.

That said, I still haven't tested with 32-bit and 64-bit together, so don't merge this yet.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1182,6 +1182,6 @@ IF ERRORLEVEL == 1 (
     pause
   )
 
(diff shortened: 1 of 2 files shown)


jb-alvarado:
I have not found a option to delete it :). I have already a bug report in the issues that the build_successful* works not right. So maybe is better to merge it and then we see how to fix that.

jb-alvarado:
Maybe we need now for the git/hg/svn sources a general check that they a already in the system or not. I had here the problem that yesterday I run the 32 bit versions, and today I wanned to run the 64 bit version, but it breaks because some libs was not in the system, because the folder was already there and up to date.

wiiaboo:
Like this? https://github.com/wiiaboo/media-autobuild_suite/commit/1d7d8cdabfad9e883187c7d62638fca4cbff0ce0

wiiaboo:
Lol, now that I'm testing both arch I realized "make uninstall" were uninstalling the 32-bit stuff even if compiling 64-bit. We have to change every one to manual uninstall, I guess?

PR #77, opened 2015-04-12
Fixes

wiiaboo:
Now that I'm testing clean build from scratch all these fixes keep coming up.

Could you add me as contributor after all? I'll try to only push small fixes without creating pull requests.


wiiaboo:
Finally, everything seems to work.

jb-alvarado:
I have add you to the list of contributors.

wiiaboo:
I'll add just the fixes then and leave the make uninstall change for later.

PR #79, opened 2015-04-13
Git check

wiiaboo:
Big fixes, so opening a pull request.

I'm adding checks for every package to prevent build fails not to fix, because apparently it works.

Will test as soon as I finish ricing every package.


wiiaboo:
Lol, can't test because waio's git is down.

jb-alvarado:
Look fine for me, when you test has work we can push it.

wiiaboo:
mpv still compiles without waio, so I should be able to test anyway.

wiiaboo:
Test seems ok. Will merge.

PR #83, opened 2015-04-14
Optional apps

wiiaboo:
These are the options that should not be optional to choose since they all take a long time to build and add a bit of size to ffmpeg (vpx, x264 and x265).

People might also not want mediainfo and sox. SoxR can still be built without Sox since it's just its resampling library.

The last two commits make x264 and x265 actually optional, since we now only choose between building binaries and libs or libs only.

The rest of the options will be behind one last option, so we can merge this now and save the rest for a later bunch of commits in a week or more so we don't delete people's INIs too much.


wiiaboo:
And we have to fix remaining bugs first if there are any, obviously.

jb-alvarado:
For me it looks good, so yes we can merge it.

PR #89, opened 2015-05-10
Build gpac and mp4box shared

wiiaboo:
We'll wait for gpac to reply before commiting this.

### DIFF
--- a/media-suite_compile.sh
+++ b/media-suite_compile.sh
@@ -1738,11 +1738,11 @@ if [[ $mp4box = "y" ]]; then
         rm -rf ./gpac-svn/
     fi
 
-    do_git "https://github.com/gpac/gpac.git" gpac noDepth master bin-video/MP4Box.exe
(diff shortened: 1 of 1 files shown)


jb-alvarado:
Have post here: https://github.com/gpac/gpac/issues/139

wiiaboo:
By the way, if you noticed, static mp4box also disables a52dec and libmad, so we could probably remove those.

jb-alvarado:
Hm, when this is only for gpac than yes we can remove them. But maybe they was for sox to. I don't remember exactly.

jb-alvarado:
It looks like, that it takes a bit long for a fix. So what you think we merge your fix?

wiiaboo:
What do you think is better? Shared or checkout to commit before and continue building static?

jb-alvarado:
I think we can build for now the shared and when it is fixed, we can change again. Ok?

PR #95, opened 2015-05-24
Remove redundant extra-libs from ffmpeg

wiiaboo:
FFmbc can be left as is since it's based on an old version of FFmpeg, so things might or not break if we take extra-libs off, but I just finished a full compilation from scratch and there were no problems.

How about on your side, @jb-alvarado ?

### DIFF
--- a/media-suite_compile.sh
+++ b/media-suite_compile.sh
@@ -1458,8 +1458,11 @@ if [ -f "$LOCALDESTDIR/lib/libxavs.a" ]; then
 
         ./configure --host=$targetHost --prefix=$LOCALDESTDIR --bindir=$LOCALDESTDIR/bin-video
 
-        make -j $cpuCount
-        make install
+        make -j $cpuCount libxavs.a
+
+        install -m 644 xavs.h $LOCALDESTDIR/include
+        install -m 644 libxavs.a $LOCALDESTDIR/lib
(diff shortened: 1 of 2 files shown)


jb-alvarado:
I just finish the 32bit compiling, and yes ffmpeg build without the extra-libs. I only had to fix libsndfile:

```
if [[ ! -f ./configure ]]; then
		sed -i "s/python/python2/" autogen.sh
        ./autogen.sh
    else
```

wiiaboo:
Oh? Even with https://github.com/jb-alvarado/media-autobuild_suite/commit/84dfc8a2a8099faa1aa6ad07b5b35950d62b060d ?
I was compiling again from zero after https://github.com/jb-alvarado/media-autobuild_suite/commit/f2877b19defce205c9b1ffa69126eb2272f2b0da and twolame failed because ffmpeg doesn't use pkg-config.

jb-alvarado:
Ah... This I have not see. I only get the shell script from you...

wiiaboo:
32-bit bit from scratch worked.

jb-alvarado:
ffmpeg 64bit without extra-libs runs here also smooth.

wiiaboo:
64-bit build from scratched worked too. I think it's safe to merge.

PR #100, opened 2015-05-31
Mkvtoolnix Qt GUI

wiiaboo:
Qt version of mkvtoolnix's GUI has come out of alpha. After 8.0.0, the wxGTK version will be considered deprecated.

Problems with this:
We need either to compile qtbase (just `make clean` takes ages) for ourselves or install `qt5-static`.
Compiling by ourselves takes a buttload of time to make and clean but we get a lighter version of qt.
Installing qt5-static we add a 500MB download and 3GB install size just from that, but I've confirmed installation works with the latter.
I haven't been able to finish the compilation of Qt yet, and I don't think it's worth it.

To prevent this 3GB behemoth from getting installed even if you aren't compiling mkvtoolnix, we could add a mechanism to only install it if mkvtoolnix is selected.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -662,8 +662,9 @@ if %mkvINI%==0 (
     echo -------------------------------------------------------------------------------
     echo.
     echo. Build static mkvtoolnix binaries?
-    echo. 1 = Yes
+    echo. 1 = Yes (Qt GUI^)
     echo. 2 = No
+    echo. 3 = Yes (wxGTK and Qt GUI^)
     echo.
     echo -------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


wiiaboo:
Managed to compile Qt with out-of-tree build. 60MB total after install, lol. Much more worth it than qt5-static's 3GB.

wiiaboo:
We can compile both wxGTK and Qt GUIs too until the wxGTK gets completely deprecated.
Or add an option to the bat like I did with x265.

wiiaboo:
Build failed with 64-bit for some reason, not ready for merge yet.

jb-alvarado:
At the begin I had compile qt4 for vlc, maybe there is some useful commands.

PR #134, opened 2015-08-21
Enable libkvazaar in FFmpeg

aryla:
These commits modify the compilation script to build a static kvazaar library and enable it when building FFmpeg.


wiiaboo:
I tried ffmpeg configure before going to bed yesterday and it was failing with libkvazaar enabled.
I'll try again today with those CFlags.

wiiaboo:
Must've been the lack of Cflag then. Works fine now, thanks!

PR #167, opened 2015-09-21
modified:   media-autobuild_suite.bat

mboncoddo:
Fix for msys installation script break due to syntax.


wiiaboo:
Can you explain why that fixes it? Does whitespace matter to batch or something?

mboncoddo:
Yeah whitespace matters in some cases. A space was needed between ( and set.

PR #236, opened 2016-01-01
Replaces the git repository for dvdread

Creased:
Since videolan git repository for dvdread seems to be unavailable, I
suggest to replace it with a mirror.


Creased:
Oops, it seems that the git repository was down just for few minutes... It works now ! But at least, I can confirm that this repository is a pure mirror.

wiiaboo:
videolan git repos should be generally stable. They're the main repos for a lot of projects.

PR #241, opened 2016-01-04
Missing ; then on line 1400

cnrd:
When you did this commit: 9c01be97407259c297240d915967901f14895aad

You missed ; then after the elif statement, which makes the compile script not run.


wiiaboo:
It's actually more work to merge this than to commit and push.

PR #440, opened 2017-01-04
Remove '--disable-w32threads' from zeranoe options

softworkz:
Latest zeranoe builds no longer use --disable-w32threads

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -69,7 +69,7 @@ set ffmpeg_options=--enable-avisynth --enable-gcrypt --enable-libmp3lame ^
 --enable-libopus --enable-libvorbis --enable-libvpx --enable-libx264 --enable-libx265 ^
 --enable-cuda --enable-cuvid --enable-schannel --enable-sdl2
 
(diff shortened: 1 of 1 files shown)


softworkz:
There is also some logic to add "disable-w32threads" in case decklink is enabled.
I'm not sure if this is required any longer because the latest zeranoe builds include decklink without specifying "disable-w32threads"

wiiaboo:
No idea. That option was there since before my time. As long as it builds and there's no users of decklink complaining it breaks something, lgtm.

softworkz:
Yeah, you're right, zeranoe's ffmpeg doesn't have decklink anymore although it is using --enable-decklink
I agree that zeranoe must decide what he  wants to do...

wiiaboo:
I'll probably still remove the --disable-w32threads since the script adds it automatically anyway.

softworkz:
I had the same thought, but that doesn't really matter a lot.

But I noticed that there are a lot of other discrepancies between your list of zeranoe options and the actual options used. It's not a big issue since you can create your own list, but I noticed that 'your' zeranoe configuration is far away from the current zeranoe build configuration...

wiiaboo:
I'm not the guy nor do I use his builds. I have no idea when he changes his options.

PR #480, opened 2017-02-22
Make media-autobuild_suite.bat executable

polymetric1:
If checked out with anything other than msysgit (Git for Windows), the
main batch file would throw an error on startup, saying the device or
file was not accessible. This permission change (`chmod +x`) fixes that.


wiiaboo:
Lol, does this even matter for Windows?
What other git client is that, btw?

polymetric1:
Yeah, it does, somehow. It was checked out with the Cygwin git. I tried with MSYS2 and it happened too. I just used Cygwin to do `chmod +x media-autobuild_suite.bat`. It's a rare occasion but it should fix this problem if anyone else checks out the suite with Cygwin/MSYS2.

wiiaboo:
Sure, not likely to break anything.

PR #548, opened 2017-06-14
fix libmysofa detection

eltoddy:
same as PR #545


wiiaboo:
Oh, that's what I force-pushed by accident.

PR #556, opened 2017-06-25
Updated leptonica source code download location and incremented version

kreischweide:
The download location for 1.74.1 throws a 404 right now, so I guess they removed the old downloads in favor of the current 1.74.4 version. Updated URL and hash accordingly. Download and extract/conf/make/install ran through so far.


wiiaboo:
Reworked the commit title to follow same pattern as the others.
Thanks.

PR #606, opened 2017-09-18
FontConfig: Fix make error from 2.12.5

laserfoxcn:
FontConfig: Fix make error from 2.12.5


wiiaboo:
I can't test it, so looking at fontconfig git I'll assume fc-blanks doesn't exist anymore.

PR #627, opened 2017-11-05
fixed the issue with Neon...

Gnarmagon:
Webdav.org denies every connection !!!


wiiaboo:
Done already in b678e2e. I'd rather not've changed mirrors.

Issue #630, opened 2017-11-09
Compilation hangs after detecting libgcrpt is up-to-date


LigH-de:
I wonder if any script is asking for a confirmation, but I see no console output because MSYS2 + mintty don't display any request until I confirmed (I remember such an issue exists, maybe with Python?).

```
        Starting 32bit compilation of all tools
Imported FFmpeg options from ffmpeg_options.txt
License: nonfree
Imported mpv options from mpv_options.txt

        Starting 32bit compilation of global tools
freetype2 2.8.1 ........................................... [Up-to-date]
fontconfig git ......................................... [Already built]
harfbuzz git ........................................... [Already built]
fribidi 0.19.7 ............................................ [Up-to-date]
libass git ............................................. [Already built]
libgcrypt 1.7.8 ........................................... [Up-to-date]
_
```
...

No. Pressing ENTER seems not to advance the progress. Entering 'y' neither.

schmidthubert:
It seems to be stuck in next block gnutls not libgcrypt. Without that block the script will use gnutls libs from a previous built. Not a nice solution at all but better than nothing for now ... :relieved:

LigH-de:
Well, just before this issue occured, I edited ffmpeg_options.txt to enable most options, but ensured to keep `--enable-openssl` commented out while having `--enable-gnutls` enabled.

Is there possibly a logic issue in the previous patch [d2fbc6d](https://github.com/jb-alvarado/media-autobuild_suite/commit/d2fbc6d83704da55fa79c0f283d8ff2e12ced856) for this case?

schmidthubert:
By looking at the gnutls block I don't see a relation to that helper patch. Just wait for Ricardo to explain and fix it. :thumbsup:

Metaphorce812:
Yeah the problem is definitely related to libssl/libressl. I got the exact same error as you, so I commented out `--enable-gnutls` and un-commented `--enable-openssl`. Then libssl/libressl successfully updated, but when MABS tried to build ffmpeg I got the error from #623 again. So then I tried re-enabling gnutls and disabling openssl in ffmpeg_options.txt, hoping that the bug wouldn't be an issue since libssl/libressl was up-to date, but I just got the same hang from this issue again. So right now I can't build ffmpeg with gnutls OR openssl enabled.

schmidthubert:
[...]

PR #641, opened 2017-11-28
Fix gnutls

Metaphorce812:
Using changes to the code proposed in the discussion for issue #630, this will fix gnutls (for now at least), stop MABS from hanging, and would allow #630 to be closed.


wiiaboo:
Thanks.

PR #683, opened 2017-12-30
Fallback patch location to build/ folder

GyanD:
wget downloads patches to the lib's subfolder within /build but the next test looks in /patches which doesn't exist. This patch strips leading path if patch isn't found in /patches


wiiaboo:
Thanks, I went with https://github.com/jb-alvarado/media-autobuild_suite/commit/4dac7b9f03282b51586ddaa478f0f76d32cddc67 instead.

PR #729, opened 2018-01-26
Switch to libopenmpt latest git release

GyanD:
Since the openmpt git repo seems to have non-linear history, use an alternative method to switch to latest release


wiiaboo:
No need, but thanks.

PR #751, opened 2018-02-16
Better compatibility with the VC compiler

kliffgomel:
It will allow to launch compilation with environment variables for the VC compiler which are set by vcvarsall.bat script.


wiiaboo:
Did you test properly to check if this doesn't break anything?

This directly reverts https://github.com/jb-alvarado/media-autobuild_suite/commit/5223be843a6ddc29ce9308d7bf49d2d1474f0c4a

kliffgomel:
Yes, I tested several times with fixes, no errors occurred.

PR #788, opened 2018-04-01
Fix -enable-vmaf in 64bit if vcs is up-to-date

### LINKED ISSUE #787: --enable-libvmaf is not used in second build
I am using --enable-libvmaf in ffmpeg_options.txt.
In the first build, --enable-libvmaf works fine.
However, in the second build, --enable-libvmaf is not used in configure of ffmpeg.
You probably need to modify media-suite_compile.sh.

schmidthubert:
This should resolve #787

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1285,15 +1285,13 @@ if enabled libxvid && [[ $standalone = y ]] && ! { files_exist "${_check[@]}" &&
 fi
 
 _check=(libvmaf.{a,h,pc})
-if [[ $ffmpeg != "no" && $bits = 64bit ]] && enabled libvmaf &&
-    do_vcs "https://github.com/Netflix/vmaf.git"; then
+if [[ $ffmpeg != "no" && $bits = 64bit ]] || do_removeOption --enable-libvmaf &&
+    enabled libvmaf && do_vcs "https://github.com/Netflix/vmaf.git"; then
     do_uninstall share/model "${_check[@]}"
     log clean make clean
     do_make INSTALL_PREFIX="$LOCALDESTDIR"
     do_makeinstall INSTALL_PREFIX="$LOCALDESTDIR"
     do_checkIfExist
-else
-    do_removeOption --enable-libvmaf
 fi
 
 _check=(ffnvcodec/nvEncodeAPI.h)


enctac:
This code is probably wrong. And I think it is complicated.
<br>
test script:
```
#!/bin/bash

ffmpeg='yes'
bits='32bit'
vmaf=true
do_vcs=true

if [[ $ffmpeg != "no" && $bits = 64bit ]] || printf 'remove --enable-libvmaf\n' &&
 $vmaf && $do_vcs; then
    printf 'clean and compile\n'
else
  printf 'do nothing.\n'
fi
```
result:
```
$ ./testScript.sh
remove --enable-libvmaf
clean and compile
```
When 32bit and do_vcs=true, compile libvmaf although it removes --enable-libvmaf ...

<br>
I Suggest below. It is clear to invalidate libvmaf when $bits=32bit, and seems to work fine.

```
_check=(libvmaf.{a,h,pc})
if [[ $bits = 32bit ]]; then
    do_removeOption --enable-libvmaf
elif [[ $ffmpeg != "no" ]] && enabled libvmaf &&
    do_vcs "https://github.com/Netflix/vmaf.git"; then
    do_uninstall share/model "${_check[@]}"
    log clean make clean
    do_make INSTALL_PREFIX="$LOCALDESTDIR"
    do_makeinstall INSTALL_PREFIX="$LOCALDESTDIR"
    do_checkIfExist
fi
```

build/media-suite_compile.sh
```diff
@@ -1285,15 +1285,13 @@ if enabled libxvid && [[ $standalone = y ]] && ! { files_exist "${_check[@]}" &&
 fi
 
 _check=(libvmaf.{a,h,pc})
-if [[ $ffmpeg != "no" && $bits = 64bit ]] && enabled libvmaf &&
-    do_vcs "https://github.com/Netflix/vmaf.git"; then
+if [[ $ffmpeg != "no" && $bits = 64bit ]] || do_removeOption --enable-libvmaf &&
```
wiiaboo: At least use `{ }`. This surely doesn't work better than the previous code.

schmidthubert: Too late now. Anyway, it works better, I had tested it several times.

PR #797, opened 2018-04-07
Add possibility for local patches

schmidthubert:
I've been using this patch for about 2 weeks now without any issue. Just recently I (had to) put the vulkan patch(es) for 1.0.68 into the patches dir and I reverted back to that version because shinchiro patches weren't up-to-date ...

Feel free to use or change it. Fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/775.

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -981,7 +981,9 @@ do_patch() {
     local strip=${3:-1}  # value of "patch" -p i.e. leading directories to strip
     [[ $patch = ${patch##*/} ]] &&
         patch="/patches/$patch"
-    do_wget -c -r -q "$patch"
+    if ! ( cp -f "/trunk$patch" $(pwd)>/dev/null 2>&1 ); then    
+        do_wget -c -r -q "$patch"
+    fi
     [[ ! -f "$patch" ]] && patch=${patch##*/}
     if [[ -f "$patch" ]]; then
         if [[ "$am" = "am" ]]; then
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -60,6 +60,9 @@ if %PROCESSOR_ARCHITECTURE%==x86 (
 set build=%instdir%\build
 if not exist %build% mkdir %build%
 
+set patches=%instdir%\patches
+if not exist %patches% mkdir %patches%
+
 set msyspackages=asciidoc autoconf automake-wrapper autogen bison diffstat dos2unix help2man ^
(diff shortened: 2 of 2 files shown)


build/media-suite_helper.sh
```diff
@@ -981,7 +981,9 @@ do_patch() {
     local strip=${3:-1}  # value of "patch" -p i.e. leading directories to strip
     [[ $patch = ${patch##*/} ]] &&
         patch="/patches/$patch"
-    do_wget -c -r -q "$patch"
+    if ! ( cp -f "/trunk$patch" $(pwd)>/dev/null 2>&1 ); then
```
wiiaboo: Why parenthesis?
Why `$(pwd)` instead of `./`?
Why in `/trunk` instead of inside `/build`?

schmidthubert: Why should the folder called something else? First of all "official patches" go in there, last is anything else. Exactly in that order. 

Second: there's no option needed. If patch cannot be copied it will wget it.

schmidthubert: >Why in /trunk instead of inside /build?
Because I've tried both and wasted a lot of time with /build. PLEASE outside!

schmidthubert: >Why $(pwd) instead of ./?

What's wrong with $(pwd)? Can you please remove your own $(pwd) usage in the script(s) first?

schmidthubert: >Why parenthesis?
Why not???

wiiaboo: > Why should the folder called be something else? First of all "official patches" go in there, last is anything else. Exactly in that order.

The "official patches" aren't there locally. They're always downloaded so they're always updated. The alternative would be keep a "patches" git dir/repo inside `build` instead.

> Because I've tried both and wasted a lot of time with /build. PLEASE outside!

What do you mean by wasting time? You could just do `/build${patches}` like you're already doing with `/trunk`. Things needed for build go in /build. There's no direct usage of `/trunk` anywhere except in this PR.

> What's wrong with $(pwd)? Can you please remove your own $(pwd) usage in the script(s) first?
> > Why parenthesis?
> Why not???

Using `$()` and `()` open a sub-shell. In your `cp` case there's absolutely no reason for that. `cp` already returns 0 or non-zero for fails. And I use `$(pwd)` when I need the absolute path, afaik. `cp` doesn't need absolute paths.

If I were to completely change your PR there wouldn't be much point to accepting it and your authorship. I still disagree with making this available since absolutely no one asked for it.

You could overwrite do_patch() inside `$HOME/custom_build_options` too, or any other function you feel isn't suiting to your usecase.

PR #798, opened 2018-04-08
fixes 2 issues

pat357:
Fixes 2 things : 

**1.** screwed_prefixes : grep -q has never any output, only a return value.
```
screwed_prefixes=($(grep -qL "$(cygpath -m "$LOCALDESTDIR")" "${screwed_prefixes[@]}"))
```
will always be empty this way.

**2. MPV and vmaf both have a a function called "write_image()"** exported to public.

This renames both read_image() and write_image() to resp. read_image2() and write_image2().
I know, just adding a 2 to the name is not the most creative thing, but it does the job !

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -81,7 +81,7 @@ _pkg_config_files=$(find "$LOCALDESTDIR/lib/pkgconfig/" -name "*.pc")
 if [[ -n "$_pkg_config_files" ]]; then
     screwed_prefixes=($(grep -E -l "(prefix|libdir|includedir)=[^/$].*" $_pkg_config_files))
     [[ -n "${screwed_prefixes[@]}" ]] &&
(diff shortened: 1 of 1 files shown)


pat357:
First of all, it is meant as a temporal workaround until vmaf has solved this in their code.

What do you mean by "on an isolated script" ??
I'm 100% sure I got all cases of write_image() replaced.
I searched the whole directory, including all extensions, and the write_image() only popped up in about 6-7 [ch] files and the rest in compiled files.
vmaf would no longer compile if I would miss something, but it compiles just fine.

It's basically the same thing as you did with the speex/codec2 IIRC, why would this be different ?

wiiaboo:
And in there I also said that codec2 devs should be notified ASAP.

wiiaboo:
I've reworked your first fix, so make a commit with just the vmaf workaround with a proper commit message and I'll merge it.

pat357:
I decided that, after all, it's better to wait for the solution from the VMAF devs.
If someone wants to compile it now, they should make their own patch or wait for VMAF.

PR #807, opened 2018-04-14
Update media-suite_compile.sh

LigH-de:
meson prefers a simpler syntax for the relative subdirectory of bindir


wiiaboo:
Didn't notice the PR was done already.

PR #808, opened 2018-04-15
Adjusting ffmpegupdate behavior

schmidthubert:
tesseract and aom/av1 are getting constant updates which should always trigger a rebuild as well ...

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1394,8 +1394,8 @@ if [[ $ffmpeg != "no" ]]; then
         _check+=(libavutil.a)
         [[ $ffmpeg = "both" ]] && _check+=(bin-video/ffmpegSHARED)
     fi
-    [[ $ffmpegUpdate = y ]] && enabled_any lib{ass,x264,x265,vpx} &&
-        _deps=(lib{ass,x264,x265,vpx}.a)
+    [[ $ffmpegUpdate = y ]] && enabled_any lib{tesseract,ass,x264,x265,vpx,aom} &&
+        _deps=(lib{tesseract,ass,x264,x265,vpx,aom}.a)
     if do_vcs "https://git.ffmpeg.org/ffmpeg.git"; then
 
         do_changeFFmpegConfig "$license"


wiiaboo:
Depends. Ideally, it'd rebuild for every external dependency, not just a static list.
tesseract is of limited usefulness to most people. Same with aom, since it's basically alpha-software.

Also, that commit title/message could be better.

schmidthubert:
>tesseract is of limited usefulness to most people

That I don't agree with. That the keyword tesseract doesn't even appear in ffmpeg -h full output doesn't say anything, check [here](https://www.google.de/search?num=100&safe=off&source=hp&ei=oDrTWre5FKnB6QS12aroBw&q=ocr+ffmpeg&oq=ocr+ffmpeg&gs_l=psy-ab.3..0i22i30k1.6207.8007.0.8205.10.8.0.2.2.0.104.741.6j2.8.0....0...1c.1.64.psy-ab..0.10.785...0j0i131k1.0.HCQ7dDWAhCk) and [here](https://www.google.de/search?num=100&safe=off&source=hp&ei=5jrTWseIMayd6ATO7broBg&q=site%3Affmpeg.org+ocr+tesseract&oq=site%3Affmpeg.org+ocr+tesseract&gs_l=psy-ab.3...1461.13552.0.13733.32.31.1.0.0.0.161.1946.29j2.31.0....0...1c.1.64.psy-ab..0.8.636...0j0i131k1.0._1rC6UoW1d8) for instance to get a better idea of what's going on with that ... 

As for AOM: I have a testdir with a batch and some own recorded files where I test the progress of AOM. But at times (weekends evening etc) ffmpeg is not updated but AOM is. And then I need to delete my ffmpeg.exe to force it to rebuild because it is ignoring AOM updates. This doesn't make any sense.

So this patch fixes ffmpeg noting these 2 important "newer dependencies" and then rebuilds, that's all. It would rebuild itself anyway if it had any own updates. So ...

As for the patch title: feel free to choose (a better) one if you accept this.

Otherwise simply close this, still "it works for me" though. :+1:

wiiaboo:
Sure, but that fixes for the libs you care about. Some time later someone says the want to refresh ffmpeg for vmaf updates. And this line just goes on to every lib possible.

schmidthubert:
The "care about" part was already there, I only extended it to what updates most. Sure you could add vmaf to it and remove ass from the list as well. Otherwise you'll have to think up something completely new to get rid of the "preferred" part here ...

schmidthubert:
Thanks for changing it! :+1:

PR #811, opened 2018-04-18
Update to apache-ant 1.10.3

mkver:
Our dear friends at apache have graciously deleted the old binary again.
Also fixes a typo regarding cuda.


wiiaboo:
Could you split the commits, if possible? I can only merge in a few hours anyway.

mkver:
Done. The reason for the second commit is that the option is called "--disable-cuda-hwaccel". If one uses --disable-cuda_hwaccel in mpv_options.txt one gets an "error: no such option: --disable-cuda_hwaccel" when building mpv.

wiiaboo:
Merged in https://github.com/jb-alvarado/media-autobuild_suite/commit/2398187977175bb2817b11a6e3c00f3d3ae072d7

PR #832, opened 2018-05-04
MPV 64bit : added cflag -DARCH_X86_64=1 to use correct "nvEncodeAPI64.dll

pat357:
When compiling MPV 64bit, you'll notice a warning that "if ARCH_X86_64" evaluates to zero (wundef).
If you look at the code from **\build\nv-codec-headers-git\include\ffnvcodec\dynlink_loader.h** you will see that this prevents using the 64 bit version from "nvEncodeAPI64.dll" and instead the 32 bit version"nvEncodeAPI.dll"will be used to compile a 64 bit program!
Obviously this can never work.
Adding the cflag -DARCH_X86_64=1 solves this problem.

```
#if defined(_WIN32) || defined(__CYGWIN__)
# define CUDA_LIBNAME "nvcuda.dll"
# define NVCUVID_LIBNAME "nvcuvid.dll"
# if ARCH_X86_64
#  define NVENC_LIBNAME "nvEncodeAPI64.dll"
# else
#  define NVENC_LIBNAME "nvEncodeAPI.dll"
# endif
#else
# define CUDA_LIBNAME "libcuda.so.1"
# define NVCUVID_LIBNAME "libnvcuvid.so.1"
# define NVENC_LIBNAME "libnvidia-encode.so.1"
#endif

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1730,6 +1730,10 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
                 mpv_cflags=("-I$(cygpath -sm "$CUDA_PATH")/include")
                 mpv_ldflags+=("-L$(cygpath -sm "$CUDA_PATH")/lib/x64")
             fi
+            mpv_cflags+=("-DARCH_X86_64=1")
+            export ARCH_X86_64=1
+        else
+            unset ARCH_X86_64
         fi
         enabled libssh && mpv_ldflags+=("-Wl,--allow-multiple-definition")
         if ! mpv_disabled manpage-build || mpv_enabled html-build; then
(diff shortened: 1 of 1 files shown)


wiiaboo:
This should be fixed upstream. Send mpv this issue.

pat357:
We might as well remove the ffnvcodec-headers from the suite as we both know by now it can never work.
I'm sure you are also against misleading the other users like I'm ; we should warn them to disable this option (because by default it's enabled).

wiiaboo:
What do you mean? It works fine for me. ffmpeg uses both nvdec and nvenc and mpv uses just nvdec.

pat357:
The related file "nvEncodeAPI64.dll" is only used for encoding ; FFmpeg figures out on his own when we're compiling 64bit and adds this c-flag automatically. If you look at the ffmpeg code, it has a lot of such "if ARCH_X86_64".
MPV could use the nvEncodeAPI  for encoding if it has the right architecture flag to see if the 32bit nvEncodeAPI or the 64 bit nvEncodeAPI64 should be called.

pat357:
BTW, I would understand your **wontfix** if there were major disadvantages to my patch, but AFAIK there are zero/zip/ null/nada disadvantages.
If you see any potential problems with it, please share them with me.

wiiaboo:
Why the hell would mpv use nvenc? It uses libavcodec/libavformat for every part relative to encoding.
Either you're not getting the point, or I'm not explaining it good enough for you.

mpv isn't mencoder. You've compiled FFmpeg with nvenc support, then mpv uses that, it doesn't recompile every fucking shit again like mencoder,

wiiaboo:
https://git.videolan.org/?p=ffmpeg/nv-codec-headers.git;a=commit;h=9301ec63771688b01abe60df55b6be5a01385fb1

There you go, the nvcodec maintainer fixed it.

pat357:
Fixed what ?  You just said that there was nothing wrong ...?
Keep your facts straight man, you look to me like an idiot atm. !

PR #836, opened 2018-05-08
Update media-autobuild_suite.bat

Strikevld:
An exception when calling "DownloadFile" with "2" arguments: "The request was interrupted: Failed to create a secure SSL / TLS channel."

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -999,7 +999,7 @@ if exist %build%\msys2-base.tar.xz GOTO unpack
 if exist %build%\wget.exe if exist %build%\7za.exe if exist %build%\grep.exe GOTO checkmsys2
 if not exist %build%\wget.exe (
     powershell -noprofile -command ^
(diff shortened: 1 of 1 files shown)


media-autobuild_suite.bat
```diff
@@ -999,7 +999,7 @@ if exist %build%\msys2-base.tar.xz GOTO unpack
 if exist %build%\wget.exe if exist %build%\7za.exe if exist %build%\grep.exe GOTO checkmsys2
 if not exist %build%\wget.exe (
     powershell -noprofile -command ^
-    "(new-object System.Net.WebClient).DownloadFile('https://i.fsbn.eu/pub/wget-pack.exe', 'wget-pack.exe')"
+    "[System.Net.ServicePointManager]::SecurityProtocol = 'Tls12';$request = New-Object System.Net.WebCLient; $request.DownloadFile('https://i.fsbn.eu/pub/wget-pack.exe', 'wget-pack.exe')"
```
wiiaboo: Can you split the lines to avoid going beyond 100 characters wide?
Also, leave the `(new-object System.Net.WebClient).` part as it was. There's no need to create a variable.

PR #852, opened 2018-06-04
media-suite_compile.sh changes

rld243:
1. I don't see why use Google's repo for glslang but KhronosGroup's for other dependencies. Google basically just pulls from KhronosGroups's repo with no changes once a week or so.

2. crossc uses SPIRV-Cross as its submodule which is always pointing to the same SPIRV-Cross's commit. Since the owner of crossc repo doesn't update his repo frequently it can potentially lead to some problems, see: https://github.com/mpv-player/mpv/issues/5760 So change it to always pull the latest SPIRV-Cross.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1708,7 +1708,7 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
             fi
         }
 
-        add_third_party "https://github.com/google/glslang.git"
+        add_third_party "https://github.com/KhronosGroup/glslang.git"
         add_third_party "https://github.com/KhronosGroup/SPIRV-Tools.git" spirv-tools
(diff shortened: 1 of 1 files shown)


wiiaboo:
Doesn't that same mpv issue have a comment from the crossc author that he updated it to fix the error?
You shouldn't really update submodules to HEAD yourself because the parent might not work with later updates.
That's why submodules point to a commit and not HEAD.
I disagree with the second commit in principle. If the crossc submodule needs to be updated you should notify crossc itself to update it and fix whatever problems.

rld243:
Well, I've been using these changes since April and never seen it failed. But okay, I see your point. Fell free to ignore this commit then.

wiiaboo:
You could still ask rossy if he's OK with replacing the sub-module with a sub-repo like is already done with shaderc, or if it's still recommended to wait for updates from him.

PR #898, opened 2018-08-18
AOM with no-tree-slp-vectorize for GCC 8.2.0

LigH-de:
To avoid aomenc failing with segmentation faults during encoding when it is built with GCC 8.2.0 (currently only in MinGW64 in the MSYS2 project), the compiler flag -fno-tree-slp-vectorize could be added for GNU C/C++ compilers. The crash is caused by the compiler producing instructions with unaligned memory access.


wiiaboo:
Wasn't there a patch being worked on?

LigH-de:
Yes ... for a week already ... not yet published. So this would only be a temporary workaround for the impatient people.

wiiaboo:
As long as you remember to remove as soon as it's fixed, sure.

LigH-de:
I will notify you to revert it when I get notified of changes in the issues I am watching.

PR #912, opened 2018-09-06
Revert "tesseract: clamp to older commit"

schmidthubert:
It's working again, compilation of lib went through without any further changes, ffpmeg built OK as well.


schmidthubert:
Needs of course manual removal of l/a/pc ..

PR #918, opened 2018-09-13
New parameters to be passed on

LigH-de:
from bat to mintty (rav1e and ripgrep)


wiiaboo:
Add them to a new line below with the `^` escape. Better than altering two lines.

LigH-de:
Alright, if that makes it nicer.

But to be postponed until rust installs...

wiiaboo:
Merged/fixed in eb7e1ae.

PR #931, opened 2018-09-26
Disable http related opus tools features

LigH-de:
Temporary workaround to avoid incompatibility issues with OpenSSL 1.1 – based on changes suggested by @slyfox1186 in https://github.com/jb-alvarado/media-autobuild_suite/issues/930#issuecomment-424733584

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -599,7 +599,7 @@ if [[ $standalone = y ]] && enabled libopus; then
     if do_vcs "https://github.com/xiph/opusfile.git"; then
         do_uninstall "${_check[@]}"
         do_autogen
-        do_separate_confmakeinstall --disable-{examples,doc}
+        do_separate_confmakeinstall --disable-{http,examples,doc}
         do_checkIfExist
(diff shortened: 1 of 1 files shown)


build/media-suite_compile.sh
```diff
@@ -608,7 +608,7 @@ if [[ $standalone = y ]] && enabled libopus; then
     if do_vcs "https://github.com/xiph/libopusenc.git"; then
         do_uninstall "${_check[@]}"
         do_autogen
-        do_separate_confmakeinstall --disable-{examples,doc}
+        do_separate_confmakeinstall --disable-{http,examples,doc}
```
wiiaboo: libopusenc doesn't have http.

LigH-de: Then treat this PR as a mistake and do what's best. I'm reaching my limits in understanding how to use github.com features...

slyfox1186: Yeah I realized that it didn't have http as well after I took some time to continue my thoughts on the matter. Still learning.

PR #937, opened 2018-09-29
Move building of rav1e into video tools chapter

LigH-de:
Cosmetic change, because rav1e belongs among the video encoders, logically.


wiiaboo:
That's not rav1e, that's rustup and rustc installation.

LigH-de:
Indeed. Scratch that.

PR #941, opened 2018-10-01
libsrt: Remove two patches, add a substitution

LigH-de:
One step forward, two steps back


wiiaboo:
Might as well remove that function instead of renaming it to something not used anywhere.
I sent a PR to fix it upstream.

LigH-de:
As you prefer, you will have more insight. So delete this PR.

PR #950, opened 2018-10-06
bat: end session after writing profiles

schmidthubert:
see https://github.com/jb-alvarado/media-autobuild_suite/issues/876

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1616,3 +1616,4 @@ for %%i in (%*) do (
     )
 setlocal
 goto :EOF
+.profile/.bash_logout


wiiaboo:
What does this do?

schmidthubert:
Because exit isn't possible then there is a different way:
https://unix.stackexchange.com/questions/423958/what-is-the-purpose-of-bash-logout/424297

wiiaboo:
Why do you expect this to work in a .bat file?

schmidthubert:
It works for me for weeks now ...

schmidthubert:
https://sourceforge.net/p/mingw/mailman/message/8685967/

wiiaboo:
The issue this is fixing never existed for me, this will just not do anything since there's no such file and it's not a valid executable for Windows even if it did exist, because .bash_logout is a bash executable script.

So again, what is this fixing and how is it fixing it?

schmidthubert:
I can't try right now. Have lots of things running/open, will try in 10min  ...

schmidthubert:
Yours doesn't work, keeps my white window open. Hmm, will remove this PR now because you don't get that behavior at all, others all the time, some just now and then, and for some people something works and for others it does not. So I don't know anymore what's going on with msys ...

PR #975, opened 2018-11-09
Documentation about Cuda compile

lars18th:
Improved the documentation for simple guides to compile with CUDA enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,10 @@ What The Individual Files Do
 `/build/ffmpeg_options.txt` & `/build/mpv_options.txt`
  - If you select the option to choose your own FFmpeg/mpv optional libraries, this file will contain options that get sent to FFmpeg/mpv's configure script before compiling. Edit this file as you wish to get a smaller FFmpeg/mpv without features you don't need or with additional features not compiled by default, if supported.
 
+
+Optional User Files
+--------
+
 `/local32|64/etc/custom_profile` & `$HOME/custom_build_options`
  - Put here any general/platform tweaks that you need for _your_ specific environment. For example, if you need to manually set the CUDA_PATH & include in the PATH the binaries for msvc and nvcc, then you can use:
 ```
--- a/README.md
+++ b/README.md
@@ -227,6 +227,7 @@ If there's some error during compilation follow these steps:
  3. If it still doesn't work, [create an issue](https://github.com/jb-alvarado/media-autobuild_suite/issues/new) and paste the URL to `logs.zip` that the script gives or attach the file yourself to the issue page.
  4. If the problem isn't reproducible by the contributors of the suite, it's probably a problem on your side. Delete /msys32, /msys64, /local32 and /local64 if they exist. /build is usually safe to keep and saves time;
  5. If the problem is reproducible, it could be a problem with the package itself or the contributors will find a way to probably make it work.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -227,6 +227,7 @@ If there's some error during compilation follow these steps:
  3. If it still doesn't work, [create an issue](https://github.com/jb-alvarado/media-autobuild_suite/issues/new) and paste the URL to `logs.zip` that the script gives or attach the file yourself to the issue page.
  4. If the problem isn't reproducible by the contributors of the suite, it's probably a problem on your side. Delete /msys32, /msys64, /local32 and /local64 if they exist. /build is usually safe to keep and saves time;
  5. If the problem is reproducible, it could be a problem with the package itself or the contributors will find a way to probably make it work.
+ 6. If you compile with `--enable-libnpp` and/or `--enable-cuda-sdk` then you need to check that: 1) The CUDA_PATH environment variable exists and points to the CUDA SDK; 2) And both `cl.exe` and `nvcc.exe` executables are available in the PATH search environment.
```
wiiaboo: The files don't have to be in PATH already, only vswhere.exe to be able to find `cl.exe`, and CUDA_PATH to be able to find `nvcc.exe`.

README.md
```diff
@@ -251,6 +252,18 @@ What The Individual Files Do
  - If you select the option to choose your own FFmpeg/mpv optional libraries, this file will contain options that get sent to FFmpeg/mpv's configure script before compiling. Edit this file as you wish to get a smaller FFmpeg/mpv without features you don't need or with additional features not compiled by default, if supported.
 
 
+Optional User Files
+--------
+
+`/local32|64/etc/custom_profile` & `$HOME/custom_build_options`
+ - Put here any general/platform tweaks that you need for _your_ specific environment. For example, if you need to manually set the CUDA_PATH & include in the PATH the binaries for msvc and nvcc, then you can use:
+```
```
wiiaboo: Add `bash` after the triple backquotes to syntax highlight the code inside.

README.md
```diff
@@ -251,6 +252,18 @@ What The Individual Files Do
...
+export CUDA_PATH=$(cygpath -sm "C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0")
+export PATH=$PATH:$(dirname "$(cygpath -u "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.15.26726\bin\Hostx64\x64\cl.exe")")
+export PATH=$PATH:$(dirname "$(cygpath -u "C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0\bin\nvcc.exe")")
```
wiiaboo: I don't like all these paths being absolute. Use variables along with some comments at least. The comments should make it clear that these will break with any update done to both the CUDA SDK and MSVS.

README.md
wiiaboo: A standard MSVS installation should come with vswhere by default, I'm pretty sure. And that's what Microsoft suggests everyone uses since VS2015.

wiiaboo: vswhere should support non-standard installs, I think that's what it was made for.

> Why? If the user put the path manually nothing more is required. You don't prefer to be simpler?

Simpler means standard, and using the official tools to get the correct paths is simpler than using guesswork or asking the user to keep track of the correct path.

wiiaboo:
Merged in 411869b8584ed19d1ea8dd40c4d9ce8a96acbff3..c8dc52c4f1eca72484275e9c46031c4a3130a138 and edited it a bit in 95e48f31408970b4285b2b36b9cfc2ba2bf348aa.

lars18th:
Hi @wiiaboo ,

:+1:

PR #977, opened 2018-11-09
First trial of NoGUI mode

lars18th:
Hi @jb-alvarado ,

This is based on discussion at https://github.com/jb-alvarado/media-autobuild_suite/issues/970

At time, unfinished. But it's a starting point: The check of remote sessions is done. Perhaps the next step can be create one script ( `mintty.bat` for example) to catch all parameters and call to `%instdir%\%msys2%\usr\bin\bash.exe` passing the correct parameters. In my first trial of this, it fails in two points: 1) no logging; 2) required to use `cygpath` to convert calls.

Any idea?

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1156,8 +1156,8 @@ if "%SSH_TTY%"=="windows-pty" (
         echo "   ******            WARNING             ******   "
         echo "   ****** console only mode (no MinTTY)! ******   "
         echo "   ******                                ******   "
-        echo "   ******    You need to manally copy    ******   "
-        echo "   ******    what's printed in a shell   ******   "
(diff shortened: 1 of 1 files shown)


1480c1:
Hey, I made a version that would possible work for you. <https://github.com/Randomderp/media-autobuild_suite/tree/batchlogging> I am unsure if it will work over ssh as I have no real way to test ssh compatibility as I do not have another computer I can have open for a long enough durration to make sure it keeps the script open. It uses powershell to log all of the outputs but it should functionally act the same. Depending on your OS, it may lack colorization. Windows does lack the orange color for console though.

lars18th:
Hi @1480c1 ,

Thank you for your fork! I'll test when I have more time. :wink:

PR #984, opened 2018-11-26
Add zstd library if tesseract is linked in ffmpeg

LigH-de:


LigH-de:
OK, will test that later and report in the issue.

PR #1016, opened 2019-01-12
update.sh : Set mintty and flex as explicit if setting base packages as deps is successful.

1480c1:
This should make it so that the update script does not ask to install even though they are already installed, just hidden.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ Included Tools And Libraries
     - curl (latest release) with WinSSL/LibreSSL/OpenSSL/GnuTLS backend
     - cyanrip (git)
     - dav1d (git)
+    - dssim (git)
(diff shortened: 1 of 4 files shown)


1480c1:
Yea, I need to learn how to properly use git very soon.

1480c1:
Finally learned how to rebase. Had to view 5 different pages to make sure I wasn't screwing up. lol

wiiaboo:
Or else, just do like most other people and just branch from this repo's master to a separate branch, only add what you want to PR and then push that branch to your fork, creating a PR from it after.
That way you don't need rebases to your personal master branch.

1480c1:
Ah! I see, I did not know I could just do that.

PR #1021, opened 2019-01-13
batch: Modified wget.ps1 to match proper syntax along with some formatting.

1480c1:
Removed unnecessary semicolon and also made sure downloadfile has an 'absolute path' along with some formatting. This will fix the problem of running the script from powershell and failing to download. (downloadfile requires an absolute path and can't handle relative if running from powershell)

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1152,9 +1152,9 @@ if exist %build%\wget.exe if exist %build%\7za.exe if exist %build%\grep.exe GOT
 setlocal enabledelayedexpansion
 if not exist %build%\wget.exe (
     (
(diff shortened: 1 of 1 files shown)


1480c1:
Seems like I lied, it seem that the command `powershell -noprofile -command "(get-filehash -algorithm sha256 wget-pack.exe).hash"` works in cmd, powershell 5, but not in powershell 6. Idk why there's this discrepancy between powershell 6 and cmd and powershell 5. I'm going to report this to powershell.

1480c1:
https://github.com/PowerShell/PowerShell/issues/8635

wiiaboo:
Sure.

PR #1024, opened 2019-01-13
Add AVS2 encoder/decoder

schmidthubert:
This is for 64bit only, there are some errors when compiling for 32bit so options are removed in that case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Included Tools And Libraries
         - decklink (10.9.3)
             - needs non-free license
         - frei0r (git)
+        - ladspa (mingw)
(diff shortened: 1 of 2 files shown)


wiiaboo:
Do people actually use these chinese MPEG clones anywhere?

schmidthubert:
I don't know. But since it was added to ffmpeg in Nov-Dec and there are examples for playing around available then I thought why not ;)

wiiaboo:
Merged in https://github.com/jb-alvarado/media-autobuild_suite/commit/4d3be234c154fdafaf54ab08734bf6d53b9cad9a

PR #1026, opened 2019-01-15
*.sh & Batch: Convert all indentions to spaces

1480c1:
Convert all 4 spaces at the beginning of the files to tab indentions. Probably not necessary, but vscode likes to do this on its own.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -3,20 +3,20 @@ shopt -s extglob
 
 FFMPEG_BASE_OPTS=(--pkg-config-flags=--static)
 if [[ x"$LOCALBUILDDIR" = "x" ]]; then
-    echo "Something went wrong."
-    echo "MSYSTEM: $MSYSTEM"
-    echo "pwd: $(cygpath -w "$(pwd)")"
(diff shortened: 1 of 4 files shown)


wiiaboo:
Your commit message says the explicit opposite of what you're doing.

Is there any specific reason other than preference for converting **to** tabs? I much prefer spaces.

1480c1:
Not particularly, I'm fine with spaces. The commit message part probably came from because I wanted to quickly commit it before my laptop died.

wiiaboo:
It could be argued that tabs are better for people that want variable length of indentation, but I also like aligning things by scope/parenthesis and for that to work you'd need to mix tabs and spaces.
It's mostly subjective anyway.

1480c1:
I see, I personally prefer tabs mainly because if I want to do a small edit on something, I usually use nano or vim as I'm usually already in a console and I don't want to open up explorer just to navigate to it to open up notepad++, and when using nano, I don't like having to hold right for every space due to the text being indented multiple times.

schmidthubert:
https://gist.github.com/eevee/6721177

PR #1036, opened 2019-01-17
update.sh: (re-)enable IgnorePkg option in /etc/pacman.conf

schmidthubert:
We shouldn't play with that option, anyone who wants to downgrade/ignore new *crap* packages for a while like currently in the zeromq case should be able to revert to a previous working package. If necessary add the list of installed packages to the error log(s) instead if this is required for problem analysis  :+1:

### DIFF
--- a/build/media-suite_update.sh
+++ b/build/media-suite_update.sh
@@ -170,7 +170,7 @@ fi
 # packet msys2 system
 # --------------------------------------------------
 
-have_updates="$(pacman -Quq)"
+have_updates="$(pacman -Qu|grep -v ignored]$|awk '{print $1}')"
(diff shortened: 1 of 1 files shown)


wiiaboo:
The suite has changed the IgnorePkg at some point. This could lead to severely out-of-date/broken packages, but sure.

schmidthubert:
Thanks! Not many will use that anyway, it's just for testing purposes and all, like glibc for vsynth etc! For MABS I currently had to downgrade my autoconf-archive package to the former 2018.03.13 because (at least) 2019.01.06 throws an error when trying to use gnutls master instead of 3_x tree ...

PR #1037, opened 2019-01-17
libzmq: fix sed

schmidthubert:
Fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/1035


wiiaboo:
Doesn't this need libsodium to be installed as well?

schmidthubert:
libsodium was in the .pc before as well and all was fine. https://github.com/zeromq/libzmq/blob/master/CMakeLists.txt mentions that "--with-libsodium" `option(WITH_LIBSODIUM "Use libsodium instead of built-in tweetnacl" OFF)` must be used  but since we don't do anything like that then I guess it's simply not needed in our case here.

PR #1038, opened 2019-01-18
Batch: Use Invoke-WebRequest instead of Net.WebClient

1480c1:
Since we are using powershell version 4 as a requirement (Based on get-filehash only being available from 4), we can use invoke-webrequest instead of the System.Net.WebClient.
Also removed algorithm arg from get-filehash since the default since its conception has been sha256.


wiiaboo:
Is that the one that's aliased to `curl`?

PR #1041, opened 2019-01-19
Batch: Added a mirror for downloading wget-pack

1480c1:
Added a mirror for wget-pack www.randomderp.com/wget-pack.exe because my school blocks all ip addresses not in the US. It will still fail properly if there is no connection to both though.

Should I add a part that would make it download from the server it has the least ping to?

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1153,11 +1153,17 @@ if exist %build%\wget.exe if exist %build%\7za.exe if exist %build%\grep.exe GOT
 setlocal enabledelayedexpansion
 if not exist %build%\wget.exe (
     (
+        echo.param^($server^)
         echo.[System.Net.ServicePointManager]::SecurityProtocol = 'Tls12'
         echo.$wc = New-Object System.Net.WebClient
-        echo.$wc.DownloadFile^('https://i.fsbn.eu/pub/wget-pack.exe', "$PWD\wget-pack.exe"^)
+        echo.if ^($server -eq "fsbn"^) {$wc.DownloadFile^('https://i.fsbn.eu/pub/wget-pack.exe', "$PWD\wget-pack.exe"^)}
+        echo.else {$wc.DownloadFile^('https://www.randomderp.com/wget-pack.exe', "$PWD\wget-pack.exe"^)}
         )>wget.ps1
-    powershell -noprofile -executionpolicy bypass .\wget.ps1
+        for /f "usebackq tokens=*" %%f in (`powershell -noprofile -command ^(Test-Connection -ComputerName i.fsbn.eu -Count 1 -InformationAction Ignore -ErrorAction Ignore^).ResponseTime`) do if not "%%f"=="" (
+            powershell -noprofile -executionpolicy bypass %build%\wget.ps1 -server fsbn
+        ) else (
+            powershell -noprofile -executionpolicy bypass %build%\wget.ps1 -server rdp
+        )
     del wget.ps1
 
     for /f "tokens=1 delims=" %%a ^


1480c1:
So, do you think it would be better to just store the exe (2MB, not much) in the build folder with the media-suite*.sh and just have it downloaded like the scripts if the build folder is deleted?
Something like:
``` batch
cd "%build%"
if not exist "%build%\wget.exe" (
    if not exist "%build%\wget-pack.exe" (
        powershell -noprofile -command Invoke-WebRequest -OutFile "$PWD\wget-pack.exe" -Uri ^
"https://github.com/jb-alvarado/media-autobuild_suite/raw/master/build/wget-pack.exe"
        for /f "tokens=1 delims=" %%a ^
in ('powershell -noprofile -command "(get-filehash -algorithm sha256 wget-pack.exe).hash"') do set _hash=%%a

    if ["!_hash!"]==["3F226318A73987227674A4FEDDE47DF07E85A48744A07C7F6CDD4F908EF28947"] (
        %build%\wget-pack.exe x
        ) else del wget-pack.exe
    )
)
```
Replacing the Invoke-WebRequest with System.Net.Webclient if prefered.

schmidthubert:
Well the blocking is politics and that's all MOSTLY rich and OLD Trump/Merkel/May/Russian/EU twats doing all this shite, younger people would never vote for such idiots as we already know by now. But I have the feeling it will take us 10-20 or more years until all these ignorant clowns get the message that we don't need more restrictions/laws/"US of Assh.. first" but less ... 

But back to the tech stuff: sure, why not put it here in the main or a tools folder? I really prefer "everything GH is down" before having to search around for a new mirror of something because someone got a divorce or decided to take his/her pages of etc. But that's not up to me to decide here ...

schmidthubert:
You can use releases:
https://help.github.com/articles/creating-releases/
https://help.github.com/articles/distributing-large-binaries/

wiiaboo:
If replacements that come OOTB in Win10 can work reasonably well (something to download msys2.tar.xz that can support resume/retries; something that can extract .tar.xz archives).
Downloaded grep is never used, so no need for that. Findstr works well enough.

schmidthubert:
Since this is "only" about a backup for the 3: for a fresh static wget 1.20.1 most people even around GH point people to https://eternallybored.org/misc/wget/ (wget.exe from there is static an works for me),  but I found this wget thread in GH more interesting as it mentions curl as an alternative (maybe you can try that instead of wget for the first steps?): 
https://github.com/luarocks/luarocks/issues/713
https://github.com/luarocks/luarocks/tree/master/win32/tools 
The wget.exe from there works for me standalone as well without any of the dlls from there even under Win7, 7za and some other binaries are there as well so there's something to try, but read below first. I haven't looked for static curl for windows now but this is it I think: https://github.com/curl/curl-for-win which has https://curl.haxx.se/download.html which then has a curl.exe inside an .xz archive, even from a very official site: https://mirrors.kernel.org/sources.redhat.com/cygwin/x86/release/curl/curl-7.59.0-1.tar.xz unfortunately not static so it complains about a cygwin dll missing but otherwise official curl for Win site is this https://curl.haxx.se/windows/ which has a curl with a dll inside.

For 7za you can find a statically linked relatively fresh exe here in GH: https://github.com/develar/7zip-bin/tree/master/win/ia32 , and for grep (findstr or not) there's this GH: https://github.com/mubix/post-exploitation/tree/master/win32bins which has a static exe in version 2.5.1 of which I don't know if it's good enough for you ... 

So for backup purposes this should be good enough and there are even some more ideas :+1:

1480c1:
For removing reliance on the downloaded wget, could I recommend
``` batch
(
        echo.[System.Net.ServicePointManager]::SecurityProtocol = 'Tls12'
        echo.$wc = New-Object System.Net.WebClient
        echo.try {
        echo.do {
        echo.$wc.DownloadFile^('http://repo.msys2.org/distrib/msys2-%msysprefix%-latest.tar.xz', "$PWD\msys2-base.tar.xz"^)
        echo.} while ^($i -le 5^)
        echo.} catch {
        echo.$i++
        echo.}
    )>msys.ps1
    powershell -noprofile -executionpolicy bypass .\msys.ps1
    del msys.ps1
```
to replace the wget in line 1193. This way we do not have to worry about wget or curl.
Since grep is also not used at all, only 7z would need to be worried about.

schmidthubert:
I don't know if this supports .xz:
https://www.powershellgallery.com/packages/7Zip4PowerShell/1.9.0
?

schmidthubert:
Why xz anyway, the exe is useless?
http://repo.msys2.org/distrib/

1480c1:
I tried that, and it works, but removing the module after using it is a pain because it uses a c# dll and you can't remove it until its completely stopped and also it seemed a bit slow.

1480c1:
The exe has no silent nor quiet as far as I can tell.

1480c1:
https://github.com/Randomderp/media-autobuild_suite/commit/aab3c6dc6af4eb32b0339ce92abf60dc070b2a68
https://github.com/Randomderp/media-autobuild_suite/commit/32caad5dc63c89db9f861bca60b979b89fefaa95
https://github.com/Randomderp/media-autobuild_suite/commit/8f8bfee926bb9c0c640127fc20cedfcaedac9899

schmidthubert:
xz for windows https://tukaani.org/xz/
don''t know if untars but https://stackoverflow.com/questions/38776137/native-tar-extraction-in-powershell ...

schmidthubert:
Is lzip not possible:
https://github.com/mcmilk/7-Zip-zstd/releases
In the releases there's an exe 
?

schmidthubert:
For 7za all that's needed is this official extra archive:
https://www.7-zip.org/a/7z1806-extra.7z
Then there's a PS module called pscx which has a native expand function:
https://www.powershellgallery.com/packages/Pscx/3.3.2
https://github.com/Pscx/Pscx
?

1480c1:
So, the pscx works and probably would be the best. I have an idea of doing
``` batch
powershell -noprofile -command {Install-Module -Name Pscx -Scope CurrentUser -Force -AllowClobber; ^
    Expand-Archive -Force $PWD\msys2-base.tar.xz; Expand-Archive -Force -OutputPath .. $PWD\msys2-base.tar}; ^
    Uninstall-Module -Name pscx -Force -AllVersions
```
in place of 7za

1480c1:
Since, this is unneeded.

PR #1042, opened 2019-01-20
Batch: Removed wget-pack.exe, replaced functions with powershell

1480c1:
Replaced wget-pack.exe functions with powershell equivalent. https://github.com/jb-alvarado/media-autobuild_suite/pull/1041
It tested working for me.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1140,48 +1140,8 @@ if %writeforceQuitBatch%==yes echo.forceQuitBatch=^%forceQuitBatchF%>>%ini%
 ::------------------------------------------------------------------
 ::download and install basic msys2 system:
 ::------------------------------------------------------------------
-if exist "%instdir%\%msys2%\usr\bin\wget.exe" GOTO getMintty
-echo -------------------------------------------------------------
-echo.
-echo - Download wget
-echo.
-echo -------------------------------------------------------------
 cd build
 if exist %build%\msys2-base.tar.xz GOTO unpack
-if exist %build%\wget.exe if exist %build%\7za.exe if exist %build%\grep.exe GOTO checkmsys2
-
-setlocal enabledelayedexpansion
-if not exist %build%\wget.exe (
-    (
-        echo.[System.Net.ServicePointManager]::SecurityProtocol = 'Tls12'
-        echo.$wc = New-Object System.Net.WebClient
-        echo.$wc.DownloadFile^('https://i.fsbn.eu/pub/wget-pack.exe', "$PWD\wget-pack.exe"^)
-        )>wget.ps1
-    powershell -noprofile -executionpolicy bypass .\wget.ps1
-    del wget.ps1
-
-    for /f "tokens=1 delims=" %%a ^
-in ('powershell -noprofile -command "(get-filehash -algorithm sha256 wget-pack.exe).hash"') do set _hash=%%a
-
-    if ["!_hash!"]==["3F226318A73987227674A4FEDDE47DF07E85A48744A07C7F6CDD4F908EF28947"] (
-        %build%\wget-pack.exe x
-        ) else del wget-pack.exe
-    )
-setlocal
-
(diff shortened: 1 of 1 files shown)


1480c1:
I see, allowclober is not on 5, back to the drawing board. Though I can implement those, just a sec.

schmidthubert:
Can you not just make your branch the master and delete everything else then? Sorry I'm not a git guru!

1480c1:
Sure!
btw, do `git clone <GIT URL> --branch <BRANCH> <DESTFOLDER>

1480c1:
Things to note so far:
- You can't remove NuGet Package provider (You would have to manually delete the things, it doesn't really affect anything)

1480c1:
Try to replace line 1180 with `echo.Install-PackageProvider -Name "NuGet" -MinimumVersion 2.8.5.201 -Force -Scope CurrentUser -Confirm:$false`
the confirm:$False should act as an automatic yes

schmidthubert:
OK, that I could have figured out by myself now having to use -Confirm:$false a lot in Exchange PS scripts now and then LOL  ;) BRB ...

schmidthubert:
Still no dice, it's still asking the same silly question, argh!?

1480c1:
hmm

1480c1:
Try to replace line 1179 and 1180 with
``` batch
echo.if ^(!^(Get-Module -Name PackageManagement^)^) {
echo.Install-Module -Name PackageManagement -Scope CurrentUser -Force -confirm:$false
```

1480c1:
I'm still setting up my windows 7 vm (can't download sp1 because microsoft.com requires a newer version of ie, but I can't download a newer ie without accessing microsoft.com

schmidthubert:
If I change line 1179+1180 like you said then the question changes back to the same long question as in my first reply here. Maybe I should upgrade to PS 6.1.2 ? But that wouldn't be the default PS install for Win7 anymore then ...

schmidthubert:
As for your download problems use chrome or FF and install User-Agent faker Addon ... :+1:

schmidthubert:
https://www.microsoft.com/en-US/download/details.aspx?id=5842

1480c1:
I think I have an idea, let me quickly code it and upload.

1480c1:
Okay, I just made it so that it will manually download the dll to the supposed correct place, based on your previous error and it should work. I think.

schmidthubert:
What I said earlier that in Win10 nuget was installed isn't true, I cannot find it. Tried your news batch there as well now but there's nothing in the assemblies directory at all, and then it errors again with the pcsx allowclobber error. So you should prepare a Win10 VM as well with no PS v6 on it but just 5.1.

Anyway, must go for now and hit the sack, need to get up early. Don't forget to make snapshots if you're playing in a VM, TTYL ... :+1:

1480c1:
Good Night, still working on it.

1480c1:
Opps

1480c1:
So far, it works on a fresh windows 8.1 installation with no modifications.

wiiaboo:
Mainstream support for Win 7 has ended in 2015 and security updates end next year, so as an end-user you shouldn't really be using it.
I'm fine with hard-requiring modern versions of Powershell, as long as you can update to them.

1480c1:
> Unfortunately for Win7 that is a ps1 in a zip and not an msu package and that ps1 cannot be run without changing the PS execution-policy first and will require admin permissions to install plus a reboot anyway

Actually, since its a package it contains microsoft's signature at the bottom
``` powershell
    # SIG # Begin signature block
    $ MIIDLwYJKoZIhvcNAQcCoII.....
```
so it doesn't need executionpolicy bypass. It will require a restart though.

1480c1:
Actually, I lied.

schmidthubert:
Yes, default is all undefined, I checked. I used unrestricted and that worked then.

PR #1046, opened 2019-01-21
Batch: Error if powershell version is less than 4

1480c1:
Added preliminary check if powershell version if less than 4.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -52,6 +52,18 @@ if not ["%instdir:~60,1%"]==[""] (
     pause
     )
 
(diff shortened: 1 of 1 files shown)


1480c1:
Hmm, let me move the logic over from batch to powershell.

1480c1:
That should fix it. I'm not exactly sure why it wasn't working before.

schmidthubert:
Yes, that works, I tried with 6 instead of 4 which returned false, which is true  :smile:

1480c1:
Added a different link that has information on how to install it along with links to the required packages instead.

PR #1049, opened 2019-01-23
Batch: Fixed the retry loop for downloading msys2

1480c1:
Properly made it so that it will try to download msys2 5 times and if the download errors out and exit straight to the error message so it doesn't download PSCX etcetera it after the fifth time.


wiiaboo:
Does DownloadFile keep the file if it failed while downloading?

1480c1:
Based on my limited testing, yes. I'm going to modify it to match my powershell script's check by checking size of tar.xz and the header from the repo.

PR #1051, opened 2019-01-24
Batch: Download msys2 if it doesn't match the header from repo

1480c1:
Will compare the file size of the tar.xz and if it doesn't match the size reported by the header from the repo, it will download, exiting with 1 if it fails 5 times. If the file doesn't exist, it will still not equal the size reported by the header.


1480c1:
wait, forgot to escape parenthesis

1480c1:
Done.

PR #1054, opened 2019-01-24
Batch: Fixed slash in minttyrc

1480c1:
I forgot that windows was backslash...


1480c1:
WAit, I realized my problem.

wiiaboo:
They should both work, no? At least I think I remember forward slashes also working.

PR #1055, opened 2019-01-24
Batch: Fixed minttyrc not being written to

1480c1:
I realized that I was still using mintty, and not just bash.


wiiaboo:
Yeah, I didn't notice it either.

PR #1061, opened 2019-01-28
Opus & shaderc

Helenerineium:
Ideally, you shouldn't need enter any text here, and your commit messages should
explain your changes sufficiently (especially why they are needed).


wiiaboo:
Much better, thanks.

PR #1065, opened 2019-01-30
Batch: Formatting for consistency

1480c1:
Formatted parenthesis to end on the same column as they had begun.
Formatted indentions to match correct indentations.
Removed unnecessary inventions.
Tried to make lines around 80-90 characters long.

I need to find out how to properly break quoted strings


wiiaboo:
Sure.

PR #1066, opened 2019-01-31
Batch: Changed one of the setlocal in writeOption to endlocal

1480c1:
Idk if this is necessary. Felt weird because it keeps on setting a local environment everytime write Option is called in the options file without ending the local environment part.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1758,5 +1758,5 @@ for %%i in (%*) do (
         echo --enable-!_opt!
     )
 )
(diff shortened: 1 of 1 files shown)


1480c1:
The endlocal should end the enabledelayedexpansion stuff as, at least based on what I've read, setlocal will be similar to a subshell in the sense that local variables will be kept in between setlocal and endlocal(or the end of a script). Since we have setlocal nested under a setlocal, it technically never goes back after expanding the variables as local vars.

PR #1067, opened 2019-02-02
Batch: Added note about Windows Def Real-time protection

1480c1:
Add a note about Windows 10 Defender's real-time protection spending ~50-60% of cpu power to scan the generated files during compilation.


1480c1:
Done. Changed the recommendation to whitelisting the directory instead of turning off Real-Time protection.

PR #1072, opened 2019-02-03
Readme: Made how to use the CUDA SDK a little more clearer

1480c1:
Added a note that CUDA is different from NVENC
Added a note of not needing custom_profile if using default
Added a link to the SDK

Should help towards people who might read it and think that the custom_profile is necessary and people who just want nvenc but though CUDA was necessary.


1480c1:
Done, however I do feel like my wording is a bit vague, but still understandable I think.

PR #1075, opened 2019-02-04
Batch: Enclose sections off so that variables don't leak to each other.

1480c1:
Tested working for compilation and update script. Information taken from <https://ss64.com/nt/endlocal.html>.
Will help with trying to run without mintty.


wiiaboo:
Awesome.

PR #1085, opened 2019-02-08
Batch & Helper & Compile: Added option for not using mintty. 

### LINKED ISSUE #970: Feature request: NoGUI option
Hi,

I don't like to put noise. Sorry for my previous issue #968.

In fact, what I like to do is to execute the `media-autobuild_suite.bat` inside a OpenSSH session (aka native SSH server in Windows 10).

After activating it, I can connect to my Windows "builder" machine. And in the CMD shell, I can execute the command `media-autobuild_suite.bat`. As you can imagine all works **until** the MinTTY is called.

So, I do this "proof of concept": I preplace inside the `media-autobuild_suite.bat` all lines like ` %mintty% --log 2>&1 %build%\firstrun.log /usr/bin/bash --login ...` with `%instdir%\%msys2%\usr\bin\bash.exe --login ...`.

With this simple change, the mintty terminal is not used at all, and the shell scripts (.sh) are called *inside* the same console (in this case the OpenSSH console), using the MinTTY's bash.exe shell in text console mode.

However, at some point, _some_ script fails (in my case, the pacman update check when it finish). But, the idea is to add support to execute the full script without "opening new GUI windows" (aka MinTTY sessions). For example calling the bat with a parameter like: `media-autobuild_suite.bat /nogui`.

What you think?

1480c1:
Closes #970 and closes #977 

Added in noMintty
Changed how mintty/powershell(bash) is called
Changed how compilation_fail works if noMintty is enabled
Added the generation of bash.ps1 if noMintty is enabled
Reset color if using noMintty

If I made any mistakes, please tell me. I'm sick rn and probably will not be able to find them.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -111,7 +111,7 @@ pdf-build libmpv-shared
 set iniOptions=msys2Arch arch license2 vpx2 x2643 x2652 other265 flac fdkaac mediainfo ^
 soxB ffmpegB2 ffmpegUpdate ffmpegChoice mp4box rtmpdump mplayer2 mpv cores deleteSource ^
 strip pack logging bmx standalone updateSuite aom faac ffmbc curl cyanrip2 redshift rav1e ^
-ripgrep dav1d forceQuitBatch vvc jq dssim avs2 timeStamp noMintty
+ripgrep dav1d vvc jq dssim avs2 timeStamp noMintty
 
 set previousOptions=0
 set msys2ArchINI=0
@@ -1108,29 +1108,6 @@ if %updateSuiteF%==2 set "updateSuite=n"
 if %updateSuiteF% GTR 2 GOTO updateSuite
 if %deleteINI%==1 echo.updateSuite=^%updateSuiteF%>>%ini%
 
-:forceQuitBatch
-if %forceQuitBatchINI%==0 (
-    echo -------------------------------------------------------------------------------
-    echo -------------------------------------------------------------------------------
-    echo.
-    echo. Force quit this batch window after launching compilation script?
-    echo. 1 = Yes
-    echo. 2 = No
-    echo.
-    echo This will forcibly close this batch window. Only use this if you have the issue
-    echo where the window doesn^'t close after launching the compilation script.
-    echo.
-    echo -------------------------------------------------------------------------------
-    echo -------------------------------------------------------------------------------
-    set /P forceQuitBatchF="Forcefully close batch: "
-) else set forceQuitBatchF=%forceQuitBatchINI%
-
-if "%forceQuitBatchF%"=="" GOTO forceQuitBatch
-if %forceQuitBatchF%==1 set "forceQuitBatch=y"
-if %forceQuitBatchF%==2 set "forceQuitBatch=n"
-if %forceQuitBatchF% GTR 2 GOTO forceQuitBatch
-if %deleteINI%==1 echo.forceQuitBatch=^%forceQuitBatchF%>>%ini%
-
 :timeStamp
 if %timeStampINI%==0 (
     echo -------------------------------------------------------------------------------
@@ -1235,7 +1212,6 @@ if %noMinttyF% GTR 2 GOTO noMintty
 if %deleteINI%==1 echo.noMintty=^%noMinttyF%>>%ini%
 
 endlocal & (
-    set forceQuitBatch=%forceQuitBatch%
     set updateSuite=%updateSuite%
     set cpuCount=%cpuCount%
     set build32=%build32%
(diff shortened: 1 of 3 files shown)


1480c1:
Tested working for both noMintty and Mintty on win8.1 from scratch. Trying as if msys2 had been set up already.

1480c1:
Things to note:

- No purple and orange, aka any colors higher than 8.

media-autobuild_suite.bat
```diff
@@ -1660,20 +1736,31 @@ if [%build64%]==[yes] (
...
+set compileArgs=--cpuCount=%cpuCount% --build32=%build32% --build64=%build64% ^
 --deleteSource=%deleteSource% --mp4box=%mp4box% --vpx=%vpx2% --x264=%x2643% --x265=%x2652% ^
 --other265=%other265% --flac=%flac% --fdkaac=%fdkaac% --mediainfo=%mediainfo% --sox=%sox% ^
 --ffmpeg=%ffmpeg% --ffmpegUpdate=%ffmpegUpdate% --ffmpegChoice=%ffmpegChoice% --mplayer=%mplayer% ^
 --mpv=%mpv% --license=%license2%  --stripping=%stripFile% --packing=%packFile% --rtmpdump=%rtmpdump% ^
 --logging=%logging% --bmx=%bmx% --standalone=%standalone% --aom=%aom% --faac=%faac% --ffmbc=%ffmbc% ^
 --curl=%curl% --cyanrip=%cyanrip% --redshift=%redshift% --rav1e=%rav1e% --ripgrep=%ripgrep% ^
---dav1d=%dav1d% --vvc=%vvc% --jq=%jq% --dssim=%dssim% --avs2=%avs2% --timeStamp=%timeStamp%'
+--dav1d=%dav1d% --vvc=%vvc% --jq=%jq% --dssim=%dssim% --avs2=%avs2% --timeStamp=%timeStamp% ^
+--noMintty=%noMintty%'
```
wiiaboo: This quote could probably be removed?

wiiaboo:
@lars18th if you could test this as well before merging, it'd be great.

media-autobuild_suite.bat
1480c1: Probably, but I left it because its been there. I don't know why, but I guess it can be removed.

1480c1:
So far, it works as intended, noMintty and Mintty.

lars18th:
> @lars18th if you could test this as well before merging, it'd be great.

I'll test it and I put here the results.
Thank you @1480c1 !

lars18th:
> @lars18th if you could test this as well before merging, it'd be great.

Test done! LGTM

lars18th:
Hi @1480c1 ,

Good job! :+1: 

Some suggestions/comments:

- I use this selections:
```

 Force quit this batch window after launching compilation script?
 1 = Yes
 2 = No

This will forcibly close this batch window. Only use this if you have the issue
where the window doesn't close after launching the compilation script.

Forcefully close batch: 2

 Are you running this script through ssh or similar?
 (Can't open another window outside of this terminal)
 1 = Yes
 2 = No

This will disable the use of mintty and print the output to this console.
There is no guarantee that this will work properly.
You must make sure that you have ssh keep-alive enabled or something similar
to screen that will allow you to run this script in the background.

SSH: 1
```

But I feel the first one (Forcefully close batch) doesn't have sense when selecting SSH. So, I suggest to interchange the order and when selecting SSH then don't ask for the other.

- It's laughly to see this at _the second line from the top_:
```
 Expanding archive...
    C:\Users\me\ffmpeg\media-autobuild_suite-ssh\build\msys2-base.tar
    [oooooooooooooooooooooooooooooooooooooooooooooooo                     ]
```
I suggest to write it at the _botton_ of the screen.

For the rest: **this patch is a must have!** :1st_place_medal:

1480c1:
For the forcequitbatch thing, its unused and basically useless atm.

For the loading bar of the expanding archive, that's from a cmdlet in powershell (think of a little powershell script) and I do not have the ability to control where is shows up, only if it shows up or not.
##### technically I could, but it would not be consistent.

wiiaboo:
Remove the force close batch option as well, and I'll merge this tomorrow.

1480c1:
Sure.

lars18th:
> For the loading bar of the expanding archive, that's from a cmdlet in powershell (think of a little powershell script) and I do not have the ability to control where is shows up, only if it shows up or not.

OK. In fact, it's a cosmetic issue.
However, you now where is the command that print this? Perhaps I can try to fix it.

Thank you all for this great improvement!

1480c1:
The command is Extract-7zarchive,  you should be able to find it nested with a echo. If you are going to try, search up using write-progress as a point but of reference.

PR #1088, opened 2019-02-12
 Recomend SSH or not

lars18th:
Auto-detect if it's an SSH session

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1138,8 +1138,13 @@ if %noMinttyINI%==0 (
     echo.
     echo. Are you running this script through ssh or similar?
     echo. ^(Can't open another window outside of this terminal^)
-    echo. 1 = Yes
-    echo. 2 = No
+    if "%SSH_TTY%"!="" (
+        echo. 1 = Yes [recommended]
+        echo. 2 = No
+    ) else (
+        echo. 1 = Yes
+        echo. 2 = No [recommended]
+    )
     echo.
     echo This will disable the use of mintty and print the output to this console.
     echo There is no guarantee that this will work properly.


wiiaboo:
I mean. If you're using the suite by SSH, it shouldn't even be an option to not use noMintty.

1480c1:
> I mean. If you're using the suite by SSH, it shouldn't even be an option to not use noMintty.

Maybe this might be a better implement

lars18th:
> I mean. If you're using the suite by SSH, it shouldn't even be an option to not use noMintty.

Hi @wiiaboo ,

With all due respect! From my point of view _everything_ should be noMintty. 

You may find it silly, and excuse me if you find it offensive, but really:
**for what do we want to use Mintty now?** :interrobang: 

As _all_ can be done using plain Powershell+Bash, Mintty's support can be removed, or not?

Please think about it with an open mind. :relaxed: 
However, close this PR if you feel it's a waste of time.

wiiaboo:
I think you've misread what I posted, or you're just assuming what you prefer is the same as what everyone prefers. It should absolutely still remain an option for everyone not on ssh. Powershell is not better than mintty as a terminal emulator, far from it.

1480c1:
There is still a couple problems with using powershell+bash for compiling.

1. The Env: can still be influenced in unexpected ways since it is called directly and not using start /I
2. The logging is not perfect and is based on if Bash can exit successfully or not
3. It's performance is based on the version installed and each version has problems on its own set of problems.
4. PS 5.1 is pretty eol in terms of update and features, and we probably will not support PS 6+ since its not guaranteed to be there.
5. Very lacking in anything relating to TTY functions, colors, etc.
6. The default codepage is not utf-8 

For using Mintty+bash
1. What happens there is pretty much tested working
2. Has good color support past 8 colors
3. Env and console is easily controlled and managed
4. If there is a bug or problem with mintty, we can expect a fix at least.
5. We've used it for a while and it works.

1480c1:
Could I recommend <https://github.com/jb-alvarado/media-autobuild_suite/commit/6a7b8b2c560b607f581098acdc442ca810035aa0> instead?
It will force the use of noMintty if over a ssh connection, else it will just continue to ask the question if the person wants to not use mintty

lars18th:
> Could I recommend [6a7b8b2](https://github.com/jb-alvarado/media-autobuild_suite/commit/6a7b8b2c560b607f581098acdc442ca810035aa0) instead?
> It will force the use of noMintty if over a ssh connection, else it will just continue to ask the question if the person wants to not use mintty

Hi @1480c1 ,

I'm not a lover of "forcing selections" in software. In fact, I prefer to leave it to the user to select whatever he wants. So, my suggestion is to **only** recommend, not enforce.

In any case, please think you on a user that he likes to use the noMintty option when not using SSH... but when using another shell that doesn't supports the GUI (for example).

1480c1:
> I'm not a lover of "forcing selections" in software. In fact, I prefer to leave it to the user to select whatever he wants. So, my suggestion is to **only** recommend, not enforce.

In this case, I feel that forcing noMintty is okay because in the first place, the user would not be able to use Mintty if over an ssh connection as this would fall under the purview of setting up the environment with the option being available in any other situation.

>In any case, please think you on a user that he likes to use the noMintty option when not using SSH... but when using another shell that doesn't supports the GUI (for example).

In the case of other shells other than ssh+cmd or ssh+PS (I'm pretty sure you cannot run batch files under ssh+bash) the user is able to pick to not use Mintty so long as SSH_TTY is not set.

lars18th:
Hi @1480c1 ,

**Please, don't do assumptions!**

The user can:
- Execute the .BAT script from a WSL or any other shell. Forget to think that PowerShell is the only shell for Windows.
- Launch the script from _any_ non-GUI shell; that's from any CONSOLE process.
- The script can be executed by a non-interactive process, so no user involved.

For sure, not all of these scenarios have sense with this project. For me the only requirement regarding this is to support SSH sessions. However, other users can have different requirements.

After saying all this, I think it's just best to leave it as it is! :wink:

wiiaboo:
So basically you agree with us that it should be forced for ssh users and be left as an option for all other cases.

PR #1091, opened 2019-02-16
Compile & Helper: Use cmake for project with cmake available.

1480c1:
Trying out this Draft Pull Request thing.
Trying out using cmake when possible.
Tested libtiff and libwebp and works so far.
Going through the rest of the packages and seeing what has a cmakelist.txt file to see if they can use cmake.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2003,7 +2003,7 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
         # fix python indentation errors from non-existant code review
         grep -ZRlP --include="*.py" '\t' third_party/spirv-tools/ | xargs -r -0 -n1 sed -i 's;\t;    ;g'
 
-        do_cmake -GNinja -DSHADERC_SKIP_TESTS=ON
+        do_cmake -DSHADERC_SKIP_TESTS=ON
         log make ninja
         cmake -E copy_directory ../libshaderc/include/shaderc "$LOCALDESTDIR/include/shaderc"
         cmake -E copy_directory ../libshaderc_util/include/libshaderc_util "$LOCALDESTDIR/include/libshaderc_util"
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -542,12 +542,11 @@ if [[ $ffmpeg != "no" || $sox = y ]]; then
     enabled libmp3lame && do_pacman_install lame
 fi
 
-_check=(ilbc.h libilbc.{{l,}a,pc})
+_check=(ilbc.h libilbc.{a,pc})
 if [[ $ffmpeg != "no" ]] && enabled libilbc && do_pkgConfig "libilbc = 2.0.3-dev" &&
     do_vcs "https://github.com/TimothyGu/libilbc.git"; then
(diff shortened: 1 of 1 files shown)


1480c1:
Supposedly, this makes it so that you can open a discussion with proposed code without it being interpreted as final code. Probably would have been useful for the no Mintty thing.

1480c1:
Probably libtiff is unnecessary, although I was having problems building webp, tesseract, and leptonica earlier using the one generated using do_pacman_install libtiff  fix_libtiff_pc. Going to try without libtiff build.

1480c1:
Doesn't seem to work with pacman libtiff.

wiiaboo:
Problem with those where cmake is optional is that usually they don't have static-only builds.
And in that point, why compile libtiff shared?

1480c1:
I'm not exactly sure what webp, leptonica, and tesseract needs from libtiff so I'm compiling it shared.

1480c1:
Do you know which one specifically did not offer static-only builds? I'll try to see if they are still non-static only or ignore those if not possible.

1480c1:
So, question, why do we use `-DUNIX=on` when there is a `-DMINGW=on`?

1480c1:
I'm seeming to have problems with tesseract, its complaining about avx right now. Need to figure out why.

wiiaboo:
> 
> 
> So, question, why do we use `-DUNIX=on` when there is a `-DMINGW=on`?

It's needed in a few projects to generate a pkg-config or use pkg-config.

1480c1:
I see, main reason why I ask is because some packages failed to build due to the DUnix flag, thinking that there were some specific headers available.

1480c1:
Next question, why do we use `https://notabug.org/RiCON/soxr.git` instead of `https://git.code.sf.net/p/soxr/code`? Is there something different between the two? Running `diff -ur` between cloned repos of the two only show stuff in the .git folder.

wiiaboo:
sf.net git was kind of unstable at the time. notabug is being used as a mirror. Same as with mxf libs.

schmidthubert:
Using cmake for sndfile (even without any params) also fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/1077

I've just tested it under Win10 Home 32bit :+1:

1480c1:
Then I will create a separate pr with the libsndfile with cmake.

build/media-suite_compile.sh
```diff
@@ -1420,7 +1420,7 @@ EOF
     fi
     }
     build_x265
-    log "install" ninja -j1 install
+    log "install" ninja install
```
wiiaboo: Ninja sometimes tries to install faster than the hard drives/file system can handle. File creation/copy should remain single-threaded when possible.

1480c1: Ah, I just realized that was an install, will fix later.

1480c1:
Right now testing with a already set up suite then going to test with a fresh one, so far it seems ok.

1480c1:
What happened to hubert? He turned into a ghost

wiiaboo:
Maybe tired or offended by something I/someone said. Not the first time
I've seen it happen in Github.

PR #1093, opened 2019-02-17
CUDA bin path added to $PATH must be in UNIX format

softworkz:
Before (wrong):
`C:/PROGRA~1/NVIDIA~2/CUDA/v8.0/bin`

After (right):
`/c/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v8.0`


wiiaboo:
Weird, worked fine for me when I did that.

PR #1095, opened 2019-02-18
Compile: Update faac to 1.29.9.2

1480c1:


wiiaboo:
faac is still being updated?? Why would anyone still maintain or use this?

1480c1:
Idk, it still seems that it's slightly active. Last commit in 2018.

PR #1102, opened 2019-02-19
gnutls: use tarball instead of git

schmidthubert:
They started [discussion](https://gitlab.com/gnutls/gnutls/merge_requests/905) again and someone requested to bring code coverage up-to-date now instead of working around it. So the previous patch cannot be applied to the 3.x stable tree, and master doesn't work yet for us here either because we are running on the latest autoconf. So using this branch is currently the only way to build gnutls successfully. We can switch later back to stable or master as you wish. I've been using master for several months now without any problems BTW, just saying. :+1:

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -289,7 +289,7 @@ fi
 _check=(libgnutls.{,l}a gnutls.pc)
 if enabled_any gnutls librtmp || [[ $rtmpdump = y ]] || [[ $curl = gnutls ]] &&
     ! files_exist "${_check[@]}" &&
-    do_wget -r -h bb9acab8af2ac430edf45faaaa4ed2c51f86e57cb57689be6701aceef4732ca7 \
+    do_wget -h bb9acab8af2ac430edf45faaaa4ed2c51f86e57cb57689be6701aceef4732ca7 \
     "https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/gnutls-3.6.6.tar.xz"; then
         do_pacman_install nettle
         do_uninstall include/gnutls "${_check[@]}"
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -289,7 +289,7 @@ fi
 _check=(libgnutls.{,l}a gnutls.pc)
 if enabled_any gnutls librtmp || [[ $rtmpdump = y ]] || [[ $curl = gnutls ]] &&
     ! files_exist "${_check[@]}" &&
-    do_wget -h bb9acab8af2ac430edf45faaaa4ed2c51f86e57cb57689be6701aceef4732ca7 \
+    do_wget -r -h bb9acab8af2ac430edf45faaaa4ed2c51f86e57cb57689be6701aceef4732ca7 \
     "https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/gnutls-3.6.6.tar.xz"; then
         do_pacman_install nettle
         do_uninstall include/gnutls "${_check[@]}"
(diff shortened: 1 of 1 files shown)


wiiaboo:
gnutls has too much activity for something that takes ages to build. We'll keep with stable releases for now.

schmidthubert:
Master doesn't work for now, as explained because of our autoconf. My remark was for later, after the coverage patches are merged. But really no problem sticking with stable at all, sure. But like I said this doesn't work either "right now" (unless reverting to previous autoconf).

1480c1:
I would say for gnutls it would be better to use tarballs because I do not consider it to be important enough to compile from Master and have to deal with every compilation problems that might come from any commits made. Might just be better to stick with a latest release and deal with problems in that release and keep it for a while until a major release or something else that might come up such as a breaking bug or security issue.

1480c1:
Mainly since it is only used in ffmpeg and curl and is optional and is not something someone might use this suite for the purpose of building gnutls.

1480c1:
Going to assume thumbs down is tarballs.

schmidthubert:
Again: whichever you choose, anything which doesn't have the coverage patches needs to use the previous autoconf. And the current autoconf will probably be outdated soon again because it also has bugs, then you'll be two versions behind with that. Is it that what you want? Please read the discussion I linked to in the first post here, all links with all details are also there. My idea was to make the gnutls tmp branch a temp solution only! I GUESS the stable gnutls branch as well as master will BOTH receive the coverage patches.

wiiaboo:
If there are tarballs in which you just need to run configure autoconf shouldn't be an issue anymore, so it's either that or that coverage branch, and I'd rather the tarball as temporary measure.

https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/gnutls-3.6.6.tar.xz

1480c1:
I would say to just use tarballs after they release one with the patch in it, eg 3.6.7, once it's available. Until then, it's probably best to change to master with the patch until they release a tarball.

schmidthubert:
Yes, thanks and sorry I wasn't even thinking about that, I was stuck in a tunnel or something. Even much better now since we got rid of that stupid bootstrap which took forever since it was always copying tons of files around.

LigH-de:
Because I was tagged with a request to reply:

> Going to assume thumbs down is tarballs.

Following @1480c1 - trusting in your possibly superior experience.

schmidthubert:
Compiling gnutls takes around 06min 30s on my main machine now, but this has quite a bunch of cores. It was taking 20-25mins before because of that slow ass bootstrap, even on this machine. The Test VM in 32bit was taking over 1h 20m hour for it with 6 cores. So this is much better. :+1:

wiiaboo:
Yeah, previously we just downloaded the newest tarballs, I can't remember why I changed to git releases.

wiiaboo:
https://github.com/jb-alvarado/media-autobuild_suite/commit/6334bdd06c8908e7307fab5bc877e6b2c045c17e maybe a partial revert of this, if we keep with tarballs instead of git.

schmidthubert:
Please do as you see fit :+1:

wiiaboo:
For now we can stay with the absolute URL. Someone just has to pay attention to a new release being done and open an issue at the time to check if we do in fact move back to tarballs or to git releases.

schmidthubert:
Yes, I will keep an eye on it as well :+1:

schmidthubert:
@wiiaboo New gnutls 3.6.7 tarball has link errors (some inet_pton redefinition problem?) but git/master continues to bootstrap/compile here just fine?

1480c1:
What was the exact error?

schmidthubert:
Sorry don't have the logs anymore and not going to try again, no time for that personally, already wasted a lot with that. It was to do with the change regarding inet_pton, please search commits/changelog/configure for this recent change if you must. It was that it cannot find the redefinition rpl_inet_pton. But in git master with bootstrap it works which technically also is version 3.6.7. You can probably get to the bottom of it by comparing 2 trees close to the release of 3.6.7 tarball, as there was no error ever with git here, just that 3.6.7 tarball does that.

schmidthubert:
https://gitlab.com/gnutls/gnutls/issues/745

This actually happens when trying to build rtmp and not gnutls itself ...

Issue #1103, opened 2019-02-21
request: x265 with SVT-HEVC


Selur:
x265 now allows to use SVT-HEVC (see: https://x265.readthedocs.io/en/latest/svthevc.html).
(no clue why they didn't make a separate project out of this)
Still it would be nice if media-autobuild_suite would also support this.

1480c1:
I'm interested in this too.
I think it's because svt-hevc is from Intel and is a separate project. afaik, you have to build svt then enable svt in x265 by flag. I think svt-hevc might require avx or sse based on the other svt-* projects.

wiiaboo:
What's the point? Why would you not just use SVT-HEVC?

1480c1:
Supposedly so that you can use x265 as the in between since x265 has more known options that they translate for you to svt and maybe be able to use svt-hevc with ffmpeg by passing options through --x265-parms(?)

1480c1:
Right now, I'm having problems with librt, seems that it's a posix, thing, going to see if there is a way around it. It seems it doesn't use avx, so on that point, maybe fine?

wiiaboo:
Try to remove -march=native if possible, since that would restrict builds of this to same or newer hardware than where it is compiled. If people want it on their `custom_build_options`, that's fine, so don't remove it from the environment's CFLAGS.

1480c1:
Will try to find a way to only run if at least sse4.1 is detected and fail if not.

<https://github.com/VcDevel/Vc/blob/master/cmake/OptimizeForArchitecture.cmake> this seems promising for at least a start.

Selur:
Any progress on this?
> What's the point? Why would you not just use SVT-HEVC?

ease of usage :) adding stv itself would be nice too :)

1480c1:
Right now, there is the code on our side to compile, problem is that even though it compiled successfully, the produced video file using the program compiled by mingw has errors.

1480c1:
https://github.com/jb-alvarado/media-autobuild_suite/pull/1105
Tentative code,
https://github.com/intel/SVT-HEVC/issues/118
Related issue on svt-hevc side

Selur:
bummer, but thanks for looking into it :)

1480c1:
Right now, mingw is not officially supported and is missing a lot of code from cmake and the c files

1480c1:
Will notify once it runs properly

Barough:
Any progress on this?

[...]

PR #1105, opened 2019-02-22
Compile: Add SVT-HEVC

1480c1:
I tried this outside of the suite, and it worked with these flags, however I would like someone else to test if it will compile properly, preferable with an older cpu. Reason why is that it sets -march=native in the cmake list and although it can be patched out, it does not change that compiling requires some extensions not present in x86_64. It seems it requires at least sse4.1 (gcc's nehalem, Intel first gen core or gcc's bdver1, AMD Bulldozer(I haven't tested on an AMD cpu yet because all of my AMD cpus are in use in linux boxes))

I does not seem that the BUILD_SHARED_LIBS=off actually do anything atm as it is hard coded .so right now. There is a pr to make it so that it can be turned off.

If I can confirm that this works, I can work on #1103 to integrate with x265.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1227,7 +1227,7 @@ _check=(SvtHevcEnc.pc libSvtHevcEnc.dll.a svt-hevc/EbApi.h
 if [[ $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
     do_uninstall bin-video/SVT-Hevc/libSvtHevcEnc.dll.a "${_check[@]}"
     do_patch "https://github.com/OpenVisualCloud/SVT-HEVC/pull/213.patch"
-    do_patch "http://0x0.st/zQKB.patch"
+    do_patch "http://0x0.st/zQP4.patch"
     do_patch "https://github.com/OpenVisualCloud/SVT-HEVC/pull/206.patch"
     do_cmakeinstall video -DUNIX=OFF
     do_checkIfExist
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -61,6 +61,7 @@ while true; do
 --avs2=* ) avs2="${1#*=}"; shift ;;
 --timeStamp=* ) timeStamp="${1#*=}"; shift ;;
 --noMintty=* ) noMintty="${1#*=}"; set >"$LOCALBUILDDIR/old.var"; shift ;;
+--svthevc=* ) svthevc="${1#*=}"; shift ;;
     -- ) shift; break ;;
     -* ) echo "Error, unknown option: '$1'."; exit 1 ;;
     * ) break ;;
@@ -1218,6 +1219,17 @@ if [[ $mp4box = "y" ]] && do_vcs "https://github.com/gpac/gpac.git"; then
     do_make
     log "install" make install-lib
     do_install bin/gcc/MP4Box.exe bin-video/
+    do_checkIfExist
+fi
+
+_check=(SvtHevcEnc.pc libSvtHevcEnc.dll.a svt-hevc/EbApi.h
+    bin-video/Svt-Hevc/{libSvtHevcEnc.dll,SvtHevcEncApp.exe})
+if [[ $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
+    do_uninstall bin-video/SVT-Hevc/libSvtHevcEnc.dll.a "${_check[@]}"
+    do_patch "https://github.com/OpenVisualCloud/SVT-HEVC/pull/213.patch"
+    do_patch "http://0x0.st/zQKB.patch"
+    do_patch "https://github.com/OpenVisualCloud/SVT-HEVC/pull/206.patch"
+    do_cmakeinstall video -DUNIX=OFF
     do_checkIfExist
 fi
 
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -113,7 +113,7 @@ pdf-build libmpv-shared
 set iniOptions=msys2Arch arch license2 vpx2 x2643 x2652 other265 flac fdkaac mediainfo ^
 soxB ffmpegB2 ffmpegUpdate ffmpegChoice mp4box rtmpdump mplayer2 mpv cores deleteSource ^
 strip pack logging bmx standalone updateSuite aom faac ffmbc curl cyanrip2 redshift rav1e ^
-ripgrep dav1d vvc jq dssim avs2 timeStamp noMintty
+ripgrep dav1d vvc jq dssim avs2 timeStamp noMintty svthevc
 
 set previousOptions=0
 set msys2ArchINI=0
@@ -412,6 +412,29 @@ if %buildother265%==1 set "other265=y"
 if %buildother265%==2 set "other265=n"
 if %buildother265% GTR 2 GOTO other265
 if %deleteINI%==1 echo.other265=^%buildother265%>>%ini%
+
+:svthevc
+if %svthevcINI%==0 (
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+    echo.
+    echo. Build SVT-Hevc? [H.265 encoder]
+    echo. 1 = Yes
+    echo. 2 = No
+    echo.
+    echo. Note: Requires at least an Intel fourth generation core or AMD Zen to
+    echo. run.
+    echo.
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+    set /P buildsvthevc="Build SVT-Hevc: "
+) else set buildsvthevc=%svthevcINI%
+
+if "%buildsvthevc%"=="" GOTO svthevc
+if %buildsvthevc%==1 set "svthevc=y"
+if %buildsvthevc%==2 set "svthevc=n"
+if %buildsvthevc% GTR 2 GOTO svthevc
+if %deleteINI%==1 echo.svthevc=^%buildsvthevc%>>%ini%
 
 :vvc
 if %vvcINI%==0 (
@@ -1561,7 +1584,7 @@ set compileArgs=--cpuCount=%cpuCount% --build32=%build32% --build64=%build64% ^
 --logging=%logging% --bmx=%bmx% --standalone=%standalone% --aom=%aom% --faac=%faac% --ffmbc=%ffmbc% ^
 --curl=%curl% --cyanrip=%cyanrip% --redshift=%redshift% --rav1e=%rav1e% --ripgrep=%ripgrep% ^
 --dav1d=%dav1d% --vvc=%vvc% --jq=%jq% --dssim=%dssim% --avs2=%avs2% --timeStamp=%timeStamp% ^
---noMintty=%noMintty%
+--noMintty=%noMintty% --svthevc=%svthevc%
     set "msys2=%msys2%"
     set "noMintty=%noMintty%"
     if %build64%==yes ( set "MSYSTEM=MINGW64" ) else set "MSYSTEM=MINGW32"


media-autobuild_suite.bat
```diff
@@ -414,6 +414,28 @@ if %buildother265%==2 set "other265=n"
...
+:svthevc
+if %svthevcINI%==0 (
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+    echo.
+    echo. Build SVT-Hevc? [H.265 encoder]
+    echo. 1 = Yes
+    echo. 2 = No
+    echo.
+    echo. Requires at least an Intel first generation core or AMD Bulldozer.
```
wiiaboo: Add note if this is required to build or to execute.

build/media-suite_compile.sh
```diff
@@ -1227,6 +1227,14 @@ if [[ $mp4box = "y" ]] && do_vcs "https://github.com/gpac/gpac.git"; then
     do_checkIfExist
 fi
 
+_check=()
+if [[ $svthevc = "y" ]] && do_vcs "https://github.com/intel/SVT-HEVC.git"; then
+    do_uninstall "${_check[@]}"
+    do_basecmake -DMINGW=on -DUNIX=off -DBUILD_SHARED_LIBS=off
```
wiiaboo: Does `do_basecmake` already exist in the suite?

build/media-suite_compile.sh
```diff
@@ -1227,8 +1227,9 @@ if [[ $mp4box = "y" ]] && do_vcs "https://github.com/gpac/gpac.git"; then
     do_checkIfExist
 fi
 
-_check=()
+_check=(SvtHevcEnc.pc libSvtHevcEnc.dll{,.a} svt-hevc/EbApi.h)
```
wiiaboo: Is this correct? Aren't you building with shared=off?

build/media-suite_compile.sh
1480c1: Crap, I forgot that its only part of my cmake branch, will fix.

wiiaboo:
Also, I assume you're going to squash the repeated commits at the end.

build/media-suite_compile.sh
1480c1: It seems that the shared=off flag doesn't do anything at all, it's hard coded to build shared libs.
There is a pr to enable the option to disable shared libs, but I will remove the dll part when they allow it, keeping it so that it's easy to spot when (Should pop up libSvtHevcEnc.dll is not found)

1480c1:
> Also, I assume you're going to squash the repeated commits at the end.

Of course.

media-autobuild_suite.bat
1480c1: Okay.

1480c1:
Right now testing as it currently is.

media-autobuild_suite.bat
```diff
@@ -423,7 +423,8 @@ if %svthevcINI%==0 (
     echo. 1 = Yes
     echo. 2 = No
     echo.
-    echo. Requires at least an Intel first generation core or AMD Bulldozer.
+    echo. Note: Requires at least an Intel first generation core or AMD Bulldozer to
+    echo. build and run. This requires sse4.1 extensions.
```
wiiaboo: Wow, even to build?

build/media-suite_compile.sh
```diff
@@ -1227,6 +1227,22 @@ if [[ $mp4box = "y" ]] && do_vcs "https://github.com/gpac/gpac.git"; then
     do_checkIfExist
 fi
 
+_check=(SvtHevcEnc.pc libSvtHevcEnc.dll{,.a} svt-hevc/EbApi.h)
+if [[ $svthevc = "y" ]] && do_vcs "https://github.com/intel/SVT-HEVC.git"; then
+    [[ $standalone = "y" ]] && _check+=(SvtHevcEncApp.exe)
+    do_uninstall "${_check[@]}"
+    ORIGINAL_PKG_CFG=$PKG_CONFIG
+    PKG_CONFIG="$LOCALDESTDIR/bin/ab-pkg-config-static.bat"
+    do_patch
```
1480c1: Ignore this for now, leaving until I can make a proper patch to replace march=native.

media-autobuild_suite.bat
1480c1: Based on my testing, yes. Although my testing may be flawed based on my knowledge with gcc.

1480c1: I could easily (Hopefully) be wrong.

1480c1:
Right now, I'm segmentation faults when running the resulting exe, will try to see if I can find out why.

1480c1:
As of right now, it seems that compiling svthevc with mingw gcc does not work and creates segmentation faults. I have submitted a pr that has a note about it, but probably don't expect it to be focused on anytime soon. Based on my limited testing, it seems that using clang might work, right now coding it, but I want to submit a different pr with <https://github.com/jb-alvarado/media-autobuild_suite/pull/1091/commits/dd4fb9181db67e0840fd4871a547067ef26cf2c7> because I would like to use pastebin for patches that I might need to edit.

build/media-suite_compile.sh
```diff
@@ -1241,6 +1241,28 @@ if [[ $mp4box = "y" ]] && do_vcs "https://github.com/gpac/gpac.git"; then
...
+    create_build_dir
+    CFLAGS="-O2 -pipe" log cmake cmake .. -G"Ninja" \
+        -DCMAKE_INSTALL_PREFIX="$LOCALDESTDIR" -DCMAKE_BUILD_TYPE=Release \
+        -DMINGW=on
+    log build ninja
+    do_install SvtHevcEnc.pc lib/pkgconfig/
+    do_install Release/lib/* lib/
+    cd_safe ..
+    do_install Bin/Release/* bin-video/SVT-Hevc/
+    do_install Bin/Release/libSvtHevcEnc.dll{,.a} lib/
```
wiiaboo: `.dll` are usually installed to the /bin-(global|video|audio) directories, since they're needed at runtime, not the compilations (directly, at least, they need them through the `.dll.a` which point to them for linking).

1480c1: I see. I will modify it after I get off this plane. However this will probably be on hold since there are still problems with the program after compilation. (The end video is split vertically and is purple)

1480c1: Although it can be compiled into x265 and ffmpeg, however right now, I'm unsure if both are actually necessary or if one is just a byproduct of improper cmake files

1480c1: If the svt-* line of encoders actually become compatible with mingw, I might just add them together

wiiaboo:
You can also use https://patchwork.ffmpeg.org/patch/12286/ (`do_patch https://patchwork.ffmpeg.org/patch/12286/mbox/ am`) to test if the lib installation is working.

1480c1:
I'll see if using the lib with ffmpeg works, maybe it's only the stand alone app that doesn't work

1480c1:
image
still no good even with using it through ffmpeg, maybe something is missing from the dll.

1480c1:
image

So currently, it just looks like a sliding motion picture

1480c1:
Although there are some issues with their output video fps and cmakelists, at the current moment, building and encoding works fine. Although it probably would be better to wait until https://github.com/OpenVisualCloud/SVT-HEVC/pull/206 and https://github.com/OpenVisualCloud/SVT-HEVC/pull/213 for bindir support, drop of march=native for default flags, and y4m container input support.

The fps issue is that the output video is marked at 25 fps even though the encoder says 29.97

wiiaboo:
If the patches are all that's needed for the build to work, you can apply them manually by the suite and then when merged upstream, the patches are removed and this can be merged.

1480c1:
Yea, that is possible. Will add

1480c1:
Actually, need to test if the prs can be combined

wiiaboo:
Does mingw not require pthreads as well?
Also, I think some projects use `find_package(Threads [REQUIRED])` to determine whether the system has pthreads through `-lpthread` or through `-pthread`, which you can retrieve from the `CMAKE_THREAD_LIBS_INIT` variable if `THREADS_FOUND` is true.

wiiaboo:
> then when merged upstream, the patches are removed and this can be merged.

I mean that this PR can be merged with the manual patches being applied right away, not only after the PRs are merged upstream, btw.

1480c1:
That's one thing I'm unsure about, although it works ATM, I do think it requires pthread

wiiaboo:
```
curl-git\CMakeLists.txt:252:  find_package(Threads REQUIRED)
aom-git\CMakeLists.txt:506:  find_package(Threads)
libwebp-git\cmake\deps.cmake:25:find_package(Threads)
shaderc-git\glslc\CMakeLists.txt:1:find_package(Threads)
shaderc-git\libshaderc\CMakeLists.txt:43:find_package(Threads)
shaderc-git\libshaderc_spvc\CMakeLists.txt:46:find_package(Threads)
shaderc-git\libshaderc_util\CMakeLists.txt:30:find_package(Threads)
aom-git\third_party\googletest\src\googletest\CMakeLists.txt:42:# Language "C" is required for find_package(Threads).
aom-git\third_party\googletest\src\googletest\cmake\internal_utils.cmake:60:    find_package(Threads)
shaderc-git\third_party\glslang\glslang\OSDependent\Unix\CMakeLists.txt:13:    find_package(Threads)
shaderc-git\third_party\glslang\glslang\OSDependent\Unix\CMakeLists.txt:18:    find_package(Threads)
```

Here's a list of projects that use them (at least that I have cloned locally, if you want to check how they do it

1480c1:
Thanks, will check them out

build/media-suite_compile.sh
```diff
@@ -1220,6 +1221,28 @@ if [[ $mp4box = "y" ]] && do_vcs "https://github.com/gpac/gpac.git"; then
     do_checkIfExist
 fi
 
+_check=(SvtHevcEnc.pc libSvtHevcEnc.dll.a svt-hevc/EbApi.h
+    bin-video/Svt-Hevc/{libSvtHevcEnc.dll,SvtHevcEncApp.exe})
+if [[ $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
+    do_uninstall bin-video/SVT-Hevc/libSvtHevcEnc.dll.a "${_check[@]}"
+    ORIGINAL_PKG_CFG=$PKG_CONFIG PKG_CONFIG="$LOCALDESTDIR/bin/ab-pkg-config-static.bat"
```
wiiaboo: Any special reason for this? You can override PKG_CONFIG in the cmake call as well without having to change the global variable.

Also, why not `do_cmake` below? Is there any broken option in it that shouldn't be called specifically for svthevc?

1480c1: At the time, I wasn't exactly sure what used the PKG_CONFIG variable, will change.

I cannot use the `-DUNIX=ON` for do_cmake since in the cmakelists for SVT-HEVC, I have some of the things filtered from APPLE, UNIX, else and some from UNIX else

1480c1: I might just bisect the do_basecmake from the other pr and add that to a separate pr so I can use that instead

wiiaboo:
Setting -DUNIX=OFF on the call to do_cmake doesn't overwrite it?

1480c1:
Maybe? Need to test extensively since I don't want to rely on something that is not reproducible

1480c1:
It works

1480c1:
Okay, done. added an extra patch in between to fix conflicts

wiiaboo:
Short and clean, nice.

Barough:
┌ SVT-HEVC git ...................................... [Recently updated]
├ Running cmake...
├ Running build...
Likely error:
                 from ../Source/Lib/C_DEFAULT/EbMcp_C.c:6:
D:/media-ab_suite/msys64/mingw32/lib/gcc/i686-w64-mingw32/9.1.0/include/emmintrin.h:143:1: error: inlining failed in call to always_inline '_mm_load_sd': target specific option mismatch
  143 | _mm_load_sd (double const *__P)
      | ^~~~~~~~~~~
In file included from ../Source/Lib/C_DEFAULT/EbMcp_C.h:9,
                 from ../Source/Lib/C_DEFAULT/EbMcp_C.c:6:
../Source/Lib/Codec/EbDefinitions.h:266:9: note: called from here
  266 |         _mm_store_sd((double*)(dst + i), _mm_load_sd((const double*)(src + i)));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ninja: build stopped: subcommand failed.
build failed. Check D:/media-ab_suite/build/SVT-HEVC-git/build-32bit/ab-suite.build.log
  Creating diagnostics file...

All relevant logs have been anonymously uploaded to https://0x0.st/ze_h.zip
Copy and paste [logs.zip](https://0x0.st/ze_h.zip) in the GitHub issue.
Make sure the suite is up-to-date before reporting an issue. It might've been fixed already.
Try running the build again at a later time.

wiiaboo:
That does not seem like a suite issue.

PR #1106, opened 2019-02-24
Helper: Add a switch to do_patch for plaintext

1480c1:
Adds a switch to not use binary mode for patches that download from places that add a cr to the lf (pastebin).

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1015,6 +1015,8 @@ do_removeOptions() {
 }
 
 do_patch() {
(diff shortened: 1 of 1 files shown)


wiiaboo:
Sure, but you shouldn't be using pastebin for anything that requires untouched text anyway.

1480c1:
Yea, eventually I plan to upload all of the pastes to 0x0, but for right now, since some of my projects require me to make patches that probably don't work the first time, I don't want to clutter 0x0 with similar but different text files.

PR #1109, opened 2019-02-26
Batch: Condense getmingws into one with Pacboy

1480c1:
We can either use pacboy or just replace the x and i with x86_64 and i386.


wiiaboo:
Sure, does pactoys-git come o.o.b. with msys?

1480c1:
I think so, I have gotten travis to work with it multiple times with a new vm. It's also included in base and base-devel.

schmidthubert:
if [%try%]==[y] (mujs

???

1480c1:
hmm, I don't know where that came from.

PR #1112, opened 2019-02-27
Update fontconfig to 2.13.1

Helenerineium:
Uses the package provided by MSYS2 instead of building from git


wiiaboo:
Did you test it with `--enable-fontconfig` and static FFmpeg? I'm pretty sure packaged fontconfig included crap that didn't work well with completely static builds last time I tried.

Helenerineium:
Yes, it was missing `-lintl` and fails at undefined reference to libintl.
Adding `-lintl` to fontconfig.pc and ffmpeg builds just fine.

wiiaboo:
What's wrong?

PR #1117, opened 2019-03-01
fontconfig: update to 2.13.1

Helenerineium:
Avoid installing dependencies since we build it from source.


Helenerineium:
Sorry about that.
I wasn't able to compile it from git (libintl)
Installing the prebulit package works but also installs unnecessary dependencies. 
Yet this pull request is again ruined by me.

PR #1118, opened 2019-03-02
fontconfig: update to 2.13.1

schmidthubert:
For a successful quick test (this will simply burn-in the current timestamp into the video with a Standard Windows font) one can run:
`ffmpeg -y -report:level=debug -i input.mp4 -vf drawtext=fontsize=24:fontcolor=yellow:x=(w-text_w)/2:y=(h-text_h)-16:fontfile='C\:\\Windows\\Fonts\\Verdana.ttf':text="TimeStamp\\:%{pts\\:hms}":box=1:boxcolor=red:boxborderw=8 -c:v h264 -c:a copy output.mp4`

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -227,7 +227,7 @@ if [[ "$mplayer" = "y" ]] || ! mpv_disabled libass ||
         if enabled libxml2; then
             do_pacman_install libxml2
             sed -i 's|Cflags:|& -DLIBXML_STATIC|' fontconfig.pc.in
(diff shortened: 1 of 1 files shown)


wiiaboo:
What's libxml2 used for?

schmidthubert:
If switched on it will be used instead of expat which is used by default, that's all that configures help gives away. You can check you bluray routine further down under `_check=(libbluray.{{l,}a,pc})`where you did something similar and switched it on as well if enabled for ffmpeg, so I did the same.

PR #1126, opened 2019-03-05
mpv: replace cuda-sdk with cuda-nvcc

schmidthubert:


wiiaboo:
Are the headers/libs still needed for cuda-nvcc?

schmidthubert:
Of course, nonsense! I changed it now! also please change commit msg if Ok for you, not time right now becauise tesseract build itself fixed but needs more fixes for ffmpeg, working on it!

PR #1127, opened 2019-03-05
tesseract: fix linking with ffmpeg

schmidthubert:
Reason for all this: https://github.com/tesseract-ocr/tesseract/commit/1c7e00611b5a6b0e2d2094f0a1326ff3c3e1fd98

Do you see any libs apart from nettle not installed with the base or should it be OK like this?

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -459,8 +459,7 @@ if [[ $ffmpeg != "no" || $standalone = y ]] && enabled libtesseract; then
         _check+=(bin-global/tesseract.exe)
         do_uninstall include/tesseract "${_check[@]}"
         sed -i -e 's|Libs.private.*|& -lstdc++|' \
-               -e 's|Libs.private.*|& -larchive -lnettle -lregex -lexpat -llzma|' \
-               -e 's|Libs.private.*|& -lzstd -llz4 -lbz2 -lz -lbcrypt -liconv|' tesseract.pc.in
+               -e 's|Requires.private.*|& libarchive iconv|' tesseract.pc.in
         do_separate_confmakeinstall global --disable-{graphics,tessdata-prefix} \
             LIBLEPT_HEADERSDIR="$LOCALDESTDIR/include" \
             LIBS="$($PKG_CONFIG --libs iconv lept libtiff-4)" --datadir="$LOCALDESTDIR/bin-global"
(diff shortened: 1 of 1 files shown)


schmidthubert:
And of course linking works now! tesseract.exe built already anyway so was easy to check what's needed. :+1:

schmidthubert:
Fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/1128

schmidthubert:
I applied this to an almost fresh script install in another VM which I just checked and the script went through. So this should be good to go.

wiiaboo:
If all those libs come from an already existing pkg-config file you could just instead use `Requires.private` instead of adding all that manually.

build/media-suite_compile.sh
```diff
@@ -454,11 +454,13 @@ if [[ $ffmpeg != "no" || $standalone = y ]] && enabled libtesseract; then
...
     if do_vcs "https://github.com/tesseract-ocr/tesseract.git"; then
-        do_pacman_install docbook-xsl
+        do_pacman_install docbook-xsl nettle
         do_autogen
         _check+=(bin-global/tesseract.exe)
         do_uninstall include/tesseract "${_check[@]}"
-        sed -i "s|Libs.private.*|& -lstdc++|" tesseract.pc.in
+        sed -i -e 's|Libs.private.*|& -lstdc++|' \
+               -e 's|Libs.private.*|& -larchive -lnettle -lregex -lexpat -llzma|' \
+               -e 's|Libs.private.*|& -lzstd -llz4 -lbz2 -lz -lbcrypt -liconv|' tesseract.pc.in
```
wiiaboo: Instead of the two lines above, do `-e 's|Requires.private.*|& libarchive iconv|'`

schmidthubert: If that would work then so should have the --libs switch, which I tried. But that complains about lots of missing *.pc files?

schmidthubert: OK, will check again.

schmidthubert: It's searching only destdir and not mingw dirs, got it, probably that's why ...

schmidthubert: On 2nd thought:  PKG_CONFIG_PATH="/local64/lib/pkgconfig:/mingw64/lib/pkgconfig" so actually it shouldn't have complained at all?

Anyway, script is running, but there are some updates ... so currently twiddling my thumbs waiting!

schmidthubert: OK, it's working, thanks for the help Ricardo! :+1:

PR #1131, opened 2019-03-11
tesseract: pacman install libarchive

schmidthubert:
I'm looking into msys2 for 32bit right now again and as I see only now that libarchive was installed if mpv is enabled/=2 . So we need to install it here as well as otherwise it would produce an error if someone didn't include mpv.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -454,7 +454,7 @@ if [[ $ffmpeg != "no" || $standalone = y ]] && enabled libtesseract; then
 
     _check=(libtesseract.{,l}a tesseract.pc)
     if do_vcs "https://github.com/tesseract-ocr/tesseract.git"; then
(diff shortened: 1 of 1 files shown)


build/media-suite_compile.sh
```diff
@@ -454,7 +454,7 @@ if [[ $ffmpeg != "no" || $standalone = y ]] && enabled libtesseract; then
 
     _check=(libtesseract.{,l}a tesseract.pc)
     if do_vcs "https://github.com/tesseract-ocr/tesseract.git"; then
-        do_pacman_install docbook-xsl nettle
+        do_pacman_install docbook-xsl nettle libarchive
```
wiiaboo: I think nettle is a libarchive dep, so you can remove it.

schmidthubert: Ah, OK, I see, "depends on". Thank you, next time I will check before!

PR #1138, opened 2019-03-17
mplayer: force enable avs demuxer

schmidthubert:
Fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/1135

`mplayer -demuxer help`


schmidthubert:
Doesn't work in 64bit, WTH is going on ...

PR #1161, opened 2019-04-09
Fix the color background

lars18th:
When chaging the background color some remote clients fail with the "dark gray" background (they set it as "black" background). So change it to "light gray".

Read about color range here: http://smallvoid.com/article/winnt-command-prompt-color.html

You can check the difference in a CONSOLE using "color 70" and "color 80".

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -21,7 +21,7 @@
 ::-----------------------------------------------------------------------------
 
 @echo off
-color 80
+color 70
 title media-autobuild_suite
 
 setlocal


lars18th:
Hi @jb-alvarado ,

When using the last version after the SSH patches (#970), I see a problem with _SOME_ SSH clients (in fact, the problem is with the SSH server you use!). Anyway the problem is when you set the gray background: in some cases it's set to BLACK Text with BLACK Background (so you don't see anything). This patch resolves this problem.

Perhaps you prefer to change it to "color 1f" or something similar. In any case, the value "70" works as expected with all the SSH clients/servers that I tested.

So please, merge the patch.

wiiaboo:
jb hasn't been in active developing here for quite some time. There's no point highlighting him.

PR #1165, opened 2019-04-12
Add Custom Patches. Closes #1163

### LINKED ISSUE #1163: User patches
Hi,

I need some help. I need to compile FFmpeg with one custom patch.

That's what I've done so far:
- Copy my patch to the directory `media-autobuild_suite-master/build/ffmpeg-git` (name like `ffmpeg-0001-my-patch.patch`).
- Edit the file `build/media-suite_compile.sh` file to add to it one line like `do_patch ffmpeg-0001-my-patch.patch` at this point:
 https://github.com/jb-alvarado/media-autobuild_suite/blob/master/build/media-suite_compile.sh#L1672

However, I suggest to incorporate some enhancement to support "user patches". For example, listing the directory `local32|64/patches` and getting patches with names `ffmpeg-*.patch`, `mediainfo-*.patch`, etc. So the script then can apply them to each project before configure/compile it.

What you think?

1480c1:
I still need to do some clean up, but here is what I have currently.

Ping @lars18th @schmidthubert if there is anything that you guys would like to change/add

I'm not sure if this is all needed, but I went with trying to add as much as directives as possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,86 @@ If there's some error during compilation follow these steps:
 
 - Put here any general/platform tweaks that you need for _your_ specific environment. See `/local32|64/etc/profile2.local` for example usage.
 
+## Custom Patches
+
+--------
+
+Using custom patches is not officially supported, if you do use custom patches, do not expect much support from this suite. Either go to the patch's author or to a forum for help
+
+- It is assumed that you, the reader, have enough knowledge of bash specifically in order to understand some basic terminology. If not, please look up what you do not know.
+- It is highly recommended to look through the [compile script](build/media-suite_compile.sh) for the exact package you want to modify and cross-reference the functions with the [helper script](build/media-suite_compile.sh) to see what is actually being done.
+- Although it is recommended to use the functions provided in the helper script, there is nothing stopping you from using the actual commands.
+
+--------
+
+- To use a custom patch, create a script within the build folder named `<repository's name>_extra.sh`.
+  - For ffmpeg, the folder name would be ffmpeg-git, but the repository's name would be ffmpeg.
+  - For game-music-emu, the folder name would be game-music-emu-0.6.2 and the repository's name would be game-music-emu
+
+- The commands will run in the cloned/unzipped folder unless otherwise stated. (flac_extra.sh would run commands within flac-git)
+
+To reference the cloned folder itself (`flac-git` `ffmpeg-git`) you can use the variable `${REPO_DIR}`.
+To reference the generated build folder, you would need to use `${REPO_DIR}/build-${bits}` to reference the build-64bit folder for 64 bit builds or build-32bit folder for 32 bit builds.
+
+- For some packages, the build folder may be slightly different (`build-shared-${bits}` for shared ffmpeg or `build-light-${bits}` for light build, etc. Try to compile at least once to see what folders are generated or look through the [compile script](build/media-suite_compile.sh))
+
+If you are building for both 32 and 64-bit, you can gate certain commands using `if [ "$bits"="64bits" ]; then <command>; fi` to only run them when building for 64 bits.\
+Be careful with certain packages due to their unique build process (x265, x264, and some others) so make sure to check the compile script
+
+Example Script: `/build/aom_extra.sh` for `aom-git`
+
+``` bash
+#!/bin/bash
+
+# Commands to run before running cmake
+_pre_cmake(){
+    # Installs libwebp
+    do_pacman_install libwebp
+    # Downloads the patch and then applies the patch
+    do_patch "https://gist.githubusercontent.com/1480c1/9fa9292afedadcea2b3a3e067e96dca2/raw/50a3ed39543d3cf21160f9ad38df45d9843d8dc5/0001-Example-patch-for-learning-purpose.patch"
+}
+
+_post_cmake(){
+    # Change directory to the build folder
+    cd_safe "build-${bits}"
+    # Rerun cmake with custom options. This will override the previous cmake commands.
+    # $LOCALDESTDIR refers to local64 or local32
+    cmake .. -G"Ninja" -DCMAKE_INSTALL_PREFIX="$LOCALDESTDIR" \
+        -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang \
+        -DBUILD_SHARED_LIBS=off -DENABLE_TOOLS=off
+}
+
+# Commands to run before building using ninja
+_pre_ninja(){
+    # Change directory to the build folder (Absolute path or relative to aom-git)
+    cd_safe "build-${bits}"
+    # applies a local patch (Absolute or relative to aom-git)
+    do_patch "/build/patches/test-diff-files.diff"
+    # run a custom ninja command.
+    ninja aom_version_check
+    # Not necessary, but just for readability sake
+    cd_safe ..
+}
+
+```
+
+Example Script: `/build/ffmpeg_extra.sh` for `ffmpeg-git`
+
+``` bash
+#!/bin/bash
+_pre_configure(){
+    # Apply a patch from ffmpeg's patchwork site.
+    do_patch "https://patchwork.ffmpeg.org/patch/12563/mbox/" am
+    # Add extra configure options to ffmpeg (ffmpeg specific)
+    # If you want to add something to ffmpeg not within the suite already
+    # you will need to install it yourself, either through pacman
+    # or compiling from source.
+    FFMPEG_OPTS+=(--enable-libsvthevc)
+}
+```
+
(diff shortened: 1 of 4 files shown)


media-autobuild_suite.bat
```diff
@@ -1624,8 +1624,8 @@ if %msys2%==msys32 (
 :: write config profiles:
 ::------------------------------------------------------------------
 
-if %build32%==yes call :writeProfile 32
-if %build64%==yes call :writeProfile 64
+if %build32%==yes if NOT EXIST %instdir%\local32\etc\profile2.local call :writeProfile 32
```
wiiaboo: Maybe move this to inside the writeProfile call?

Also, I'd prefer not having this check, since if we do need to change the contents of profile2.local we'd need to change its name again, but maybe that's the cleaner/more efficient way.

1480c1: The main reason why I had added this was because if someone wanted to change the environment variables, such  as using clang instead of GCC, through the profile2.local, it would get overwritten.

Maybe instead use something like profile.d and just drop a text file with personalize stuff to override profile2? What do you think?

build/media-suite_helper.sh
```diff
@@ -1795,3 +1829,43 @@ verify_cuda_deps() {
     fi
     enabled_any libnpp cuda-nvcc
 }
+
+check_custom_patches(){
+    local vcsFolder
+    local vcsType
+    IFS=- read vcsFolder vcsType <<< "$PWD"
```
wiiaboo: Looks cute. What does this do? Something like a `vcsFolder, vcsType = $PWD.split('-')`?
How does it look with a vcsFolder with hyphens?

media-autobuild_suite.bat
1480c1: nvm.
```


/local32|64/etc/custom_profile & $HOME/custom_build_options

    Put here any general/platform tweaks that you need for your specific environment. See /local32|64/etc/profile2.local for example usage.
```

build/media-suite_helper.sh
1480c1: Something like that, this works on the assumption that the current folder is made through do_vcs and thus has ${Repo}-${git|hg|svn}

I actually don't need the vcsType part, but just in case

1480c1: I actually did not know you can assign variables in an array like that for powershell. Thanks for showing that.

build/media-suite_helper.sh
```diff
@@ -1795,3 +1829,43 @@ verify_cuda_deps() {
...
+
+extra_script(){
+    local stage="$1"
+    local commandname="$2"
+    if type _${stage}_${commandname} >/dev/null 2>&1; then
+        log "${stage}_${commandname}" _${stage}_${commandname}
+    fi
+}
+
+unset_extra_script(){
```
1480c1: A little tempted to do this in a loop/array. However, that would make it a little unreadable to those who might not know exactly what they are.

wiiaboo: Should be fine. You can do single-line comments for each or actual documentation for them if you loop them, but in the end you should minimally document at which points they're actually run.

build/media-suite_helper.sh
1480c1: Do you know of anything better/more resilient?

wiiaboo: I'd use
```base
local _basedir="$(get_first_subdir)"
vcsFolder=${_basedir%-*}
vcsType=${_basedir##*-}
```, like I've used plenty through the scripts.

Comment (wiiaboo) [reply build/media-suite_helper.sh]: I didn't even know that was valid powershell. I knew that way of tuple assignment from python and node.

Comment (1480c1) [reply build/media-suite_helper.sh]: I realized that this is actually bad, mainly because it doesn't actually return the folder name. wait a bit

Comment (1480c1) [reply build/media-suite_helper.sh]: Maybe I should first reverse it then split?

Comment (1480c1) [reply build/media-suite_helper.sh]: Might be better for those downloaded from Sourceforge since they seem to have ${package}-${ver} generally

Comment (1480c1) [reply build/media-suite_helper.sh]: I had a problem when trying to use get_first_subdir before (was only giving me .) But I'll try again once I get home)

File: README.md [hunk 4]
```diff
@@ -270,6 +270,31 @@ If there's some error during compilation follow these steps:
...
+
+To use a custom patch, create a script within the build folder named `${REPO_NAME}_extra.sh` where `${REPO_NAME}` is something like "ffmpeg" or "x264", whatever the folder's name would be.
+
+Example Script: `/build/aom_extra.sh` for `aom-git`
+
+``` bash
+#!/bin/bash
+_pre_cmake(){
+    do_pacman_install libwebp
+    do_patch "https://gist.githubusercontent.com/1480c1/9fa9292afedadcea2b3a3e067e96dca2/raw/50a3ed39543d3cf21160f9ad38df45d9843d8dc5/0001-Example-patch-for-learning-purpose.patch" am
```
lars18th: Please, put another example using a local file.
Where targets the "relative" path if I use `do_patch "my-patch.diff" am` ?
And why the `am`?

1480c1: Sure, I guess I should specify the relative dir would be the cloned/untared folder and the am just means use git am for patch files( git apply but has commit information such as author)

wiiaboo: `am` uses `git am` to apply the patch, instead of `diff`.

build/media-suite_helper.sh
wiiaboo: Yeah, non-vcs folders are a bit of a gamble. You might use the same "algorithm" that the suite uses to create the folder for those or pass it to `check_custom_patches` instead for those.

README.md
```diff
@@ -293,7 +293,8 @@ To reference the generated build folder, you would need to use `${REPO_DIR}/buil
 
 - For some packages, the build folder may be slightly different (`build-shared-${bits}` for shared ffmpeg or `build-light-${bits}` for light build, etc. Try to compile at least once to see what folders are generated or look through the [compile script](build/media-suite_compile.sh))
 
-If you are building for both 32 and 64-bit, you can gate certain commands using `if [ "$bits"="64bits" ]; then <command>; fi` to only run them when building for 64 bits.
+If you are building for both 32 and 64-bit, you can gate certain commands using `if [ "$bits"="64bits" ]; then <command>; fi` to only run them when building for 64 bits.\
+Becareful with certain packages due to their unique build process (x265 and some others) so make sure to check the compile script
```
wiiaboo: Be careful*

wiiaboo:
Overall, looking very good. Haven't tested yet, though.

README.md
```diff
@@ -270,6 +270,86 @@ If there's some error during compilation follow these steps:
...
+    cmake .. -G"Ninja" -DCMAKE_INSTALL_PREFIX="$LOCALDESTDIR" \
+        -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang \
+        -DBUILD_SHARED_LIBS=off -DENABLE_TOOLS=off
+}
+
+# Commands to run before building using ninja
+_pre_ninja(){
+    # Change directory to the build folder
+    cd_safe "build-${bits}"
+    # applies a local patch
```
lars18th: I suggest to change to "# Applies to a local file (absolute path)".

And in the previous comment "# Change to the directory of the build folder (relative path)".

It's necessary to clarify: how to use absolute and relative pathes... and what's the root of the relative path (in fact where the file `media-autobuild_suite.bat` lives.

Good job!

wiiaboo: > how to use absolute and relative pathes... 

We're not teaching people how to use bash here. That's out of scope for the suite.

lars18th: Sorry @wiiaboo !

The idea is to provide _one_ example. Without this info we don't understand where is the root relative path when we're writing these patch scripts.

Please excuse me if I have offended you in any way, or what I have said seems foolish to you.

wiiaboo: I'm not offended. I'm just telling that we won't be teaching about absolute or relative paths.
The PWD will depend on which points you're at in the build and what package you're building. Most packages will spend most of the time inside the build-${bits}bits directory, but not for FFmpeg, mpv, etc. You can't document that to exhaustion without repeating yourself.

lars18th: Hi @wiiaboo ,

OK. But my assumption (sorry if I'm wrong) is that _all_ patch scripts (for the @1480c1 extension) will have the same syntax. So, in this case _all_ relative pathes point to the same directory, and they not be specific for each project. I'm correct about this assumption?

1480c1: They will be specific to the project.
For flac_extra.sh, they will run in flac-git
For aom_extra.sh, they will run in aom-git

lars18th:
Hi,

This PR is ready to merge?
LGTM

Issue #1147, opened 2019-03-26
vulkan-loader: TypeError in _compile() in python3.7\re.py


LigH-de:
Several header files don't compile:
```
...
  File "E:/MABS/local32/share/vulkan/registry\reg.py", line 943, in apiGen
    regEmitExtensions = re.compile(self.genOpts.emitExtensions)
  File "E:/MABS/msys64/mingw32/lib/python3.7\re.py", line 234, in compile
    return _compile(pattern, flags)
  File "E:/MABS/msys64/mingw32/lib/python3.7\re.py", line 285, in _compile
    raise TypeError("first argument must be string or compiled pattern")
TypeError: first argument must be string or compiled pattern
```
Seems to be a quite generic Python syntax issue.

[logs.zip](https://0x0.st/zKy5.zip)

LigH-de:
^ apparently related to a recent Vulkan Headers update.

LigH-de:
Comment by @jzulauf-lunarg
> Vulkan-Loader is not guaranteed to function with the tip of master on any of the constituent projects. Breakage and errors on the tip of those components aside, there may be incompatibilities, new features, or deprecations that affect compatibility. Updating to newer stable component releases is an on-going maintenance task.
So ... lock Vulkan headers to a stable release if you need Vulkan-Loader?

1480c1:
Maybe that might be a good idea.

jzulauf-lunarg:
Both Vulkan-Loader and Vulkan-ValidationLayers use "known good" versions of constituent components -- 

Vulkan-ValidationLayers/scripts/known_good.json specifies a consistent set including loader, header, and glslang, and spriv components for those building their own, with helper scripts to configure and build.

See -- [Vulkan-ValidationLayers/BUILD.md](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/BUILD.md)

jzulauf-lunarg:
Additionally, each release of the Khronos Vulkan SDK has a [branch](https://github.com/KhronosGroup/Vulkan-ValidationLayers/branches/all?utf8=%E2%9C%93&query=sdk) which could be used as a stable point for all components within the known-good list.

weezzelboy:
How do you disable this in media_autobuild_suite? I dont think i need this, but not sure i just use this for a compiled version of ffmpeg to transcode my plex library as new content come is from the automation scripts.
any help would be great.

1480c1:
[...]

PR #1167, opened 2019-04-14
Vulkan-Headers: Use known_good.json for Vulkan-Headers

1480c1:
Currently testing it.

#1147

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1822,7 +1822,7 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
     _check=(mujs.h libmujs.a)
     if ! mpv_disabled javascript &&
         do_vcs http://git.ghostscript.com/user/tor/mujs.git; then
-        do_uninstall bin-global/mujs.exe "$_{check[@]}"
+        do_uninstall bin-global/mujs.exe "${_check[@]}"
         log clean make clean
         do_make install-static prefix="$LOCALDESTDIR" bindir="$LOCALDESTDIR/bin-global"
         do_checkIfExist
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -170,10 +170,18 @@ if [[ $ripgrep = y ]] &&
     do_checkIfExist
 fi
 
+_check=(libonig.a oniguruma.pc)
+if [[ "$jq" = y ]] &&
+    do_vcs "https://github.com/kkos/oniguruma.git"; then
+    do_pacman_remove oniguruma
+    do_uninstall include/onig{uruma,gnu,posix}.h "${_check[@]}"
+    do_cmakeinstall
+    do_checkIfExist
+fi
+
 _check=(bin-global/jq.exe)
 if [[ "$jq" = y ]] &&
     do_vcs "https://github.com/stedolan/jq.git"; then
-    do_pacman_install oniguruma
     do_uninstall "${_check[@]}"
     do_autoreconf
     do_separate_confmakeinstall global --enable-all-static --enable-pthread-tls --disable-docs
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1847,13 +1847,20 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
     fi
 
     if ! mpv_disabled vulkan; then
+        # Get the known_good json and get the commit version of the Vulkan-Headers.
+        # It is assumed that the version number has quotes around them (sed)
+        [[ "$jq" != y ]] && do_pacman_install jq
+        _Vulkan_Version=$(
+            ${curl_opts[@]} https://raw.githubusercontent.com/KhronosGroup/Vulkan-ValidationLayers/master/scripts/known_good.json |
+            jq -r '.repos[] | select(.name == "Vulkan-Headers").commit')
+        [[ "$jq" != y ]] && do_pacman_remove jq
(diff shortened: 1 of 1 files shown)


build/media-suite_compile.sh
```diff
@@ -1825,13 +1825,21 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
     fi
 
     if ! mpv_disabled vulkan; then
+        # Get the known_good json and get the commit version of the Vulkan-Headers.
+        # It is assumed that the version number has quotes around them (sed)
+        [[ "$jq" != y ]] && do_pacman_install jq
+        _Vulkan_Version=$(
+            ${curl_opts[@]} https://raw.githubusercontent.com/KhronosGroup/Vulkan-ValidationLayers/master/scripts/known_good.json |
+            jq  '.repos[1].commit' |
```
wiiaboo: You can use `-r` to output raw strings.

1480c1: Oh.

build/media-suite_compile.sh
```diff
@@ -1825,13 +1825,21 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
+            sed -e 's/^"//;s/"$//')
```
```suggestion
```

build/media-suite_compile.sh
```suggestion
            jq -r '.repos[1].commit')
```

1480c1: I wasn't sure what it meant by raw instead of json

1480c1:
Worked for me

build/media-suite_compile.sh
wiiaboo: `jq --raw-output '.repos[] | select(.name == "Vulkan-Headers").commit'` should supposedly work, not apparently not in Windows.

wiiaboo:
Oh, but you're still using array index, not the actual name of the repo. I haven't managed to make it work with `select()`. But this seems to be an issue with the suite's compilation of jq. It works with the official binaries, for some reason.

1480c1:
Hmm, I have also been trying to see if I could use the name instead of a hard coded index, but all of the commands lead to `not defined at <top-level>`

1480c1:
But it seems to work with the pacman installed one

1480c1:
could this be necessary? https://github.com/msys2/MINGW-packages/blob/854b9fc6cf46206a5bb74095905ef0f4cba388a4/mingw-w64-jq/no-undefined.patch

wiiaboo:
That just fixes compiling issues, not runtime, I think.

1480c1:
I see. Hmmmmmmmmm

wiiaboo:
Ok, I've tried using `--with-oniguruma=builtin` and it seemed to fix the issue with select. I guess mingw's oniguruma is broken.

1480c1:
Right now also compiling with built-in, but if that is so, then I will change it to use a built in oniguruma according to their readme

1480c1:
Doesn't work for me it seems

1480c1:
Very tempted to just use powershell

1480c1:
`($(Invoke-WebRequest https://raw.githubusercontent.com/KhronosGroup/Vulkan-ValidationLayers/master/scripts/known_good.json | ConvertFrom-Json).repos | Where-Object {$_.name -eq "Vulkan-Headers"}).commit`

wiiaboo:
> Right now also compiling with built-in, but if that is so, then I will change it to use a built in oniguruma according to their readme

Yeah, but you need to either fix compiling it with the separate_conf or switch to compiling in-root.
Or compiling it before jq as a dependency. jq is overall more useful if we just use built-in or our own oniguruma instead of mingw's.

1480c1:
I'll try to see if we can build oniguruma and then jq instead of using built in oniguruma or pacman

1480c1:
I can confirm that it works if building oniguruma, jq, and vulkan-* from scratch

1480c1:
I will squash tomorrow morning after the win10 vm finishes. (win8 finished earlier due to the rest already being compiled)

build/media-suite_compile.sh
```diff
@@ -1825,13 +1833,20 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
+        # Get the known_good json and get the commit version of the Vulkan-Headers.
+        # It is assumed that the version number has quotes around them (sed)
+        [[ "$jq" != y ]] && do_pacman_install jq
+        _Vulkan_Version=$(
+            ${curl_opts[@]} https://raw.githubusercontent.com/KhronosGroup/Vulkan-ValidationLayers/master/scripts/known_good.json |
+            jq -r '.repos[] | select(.name == "Vulkan-Headers").commit')
+        [[ "$jq" != y ]] && do_pacman_remove jq
         _check=(vulkan/vulkan.h)
-        if do_vcs "https://github.com/KhronosGroup/Vulkan-Headers.git#tag=v1.1.102" vulkan-headers; then
+        if do_vcs "https://github.com/KhronosGroup/Vulkan-Headers.git#tag=${_Vulkan_Version}" vulkan-headers; then
```
```suggestion
        if do_vcs "https://github.com/KhronosGroup/Vulkan-Headers.git#tag=${_Vulkan_Version:-v1.1.106}" vulkan-headers; then
```

wiiaboo:
The oniguruma adding could be separated into a commit by itself for easier reverting.
The "Add default version" commit could also be squashed into the Vulkan-Headers commit, since it's a change to code added there that isn't merged yet.

1480c1:
Sure

wiiaboo:
Cool, thanks.

PR #1170, opened 2019-04-17
Patches: add verbosity to notifying use of patches

1480c1:
Also fix(?) IFS not separating on newline. 
<https://github.com/koalaman/shellcheck/wiki/SC2141>
It seems with `IFS='\n'` will match `\` or `n` and not newline

It seems that the output is the same regardless, although the second lacks an extra newline at the end of the file


wiiaboo:
Merged to 8896e77.

PR #1171, opened 2019-04-20
Use luajit instead of luajit-git

Helenerineium:


schmidthubert:
But it's called luajit-git, run 

`pacman -Ss luajit`

to see what is what?
?

schmidthubert:
Oh, IC. Fired up a test-VM and getting a 2.0.5 as luajit update as of Apr 19th there? So seems someone decided to rename that even though 2.0.4 is 2 years old and 2.0.5 is ....... [2 years old as well](http://luajit.org/) but has been renamed just now? LOL, not a single day passes here without pissing myself laughing about certain "changes" ...

1480c1:
They just changed it from using git head to using tarballs to compile

wiiaboo:
2.0.4 was a few years already.

schmidthubert:
So was 2.0.5 ? And so was 2.1.0 beta, all from 2017 ...

PR #1183, opened 2019-04-30
Repository links

1480c1:
- Switch to using github/gitlab mirrors for most non-cdn supported repositories (xiph, etc)
- Unbind ripgrep since there is now 11.0.1
- Switch to github ffmpeg mirrors
- Switch to videolan's new gitlab based code.videolan.org links
- Formatting: Add quotes around repo links and patch links
- Use hourly gitlab mirrors for notabug repos, or.cz, and fraunhofer's
  - Mainly because my win10 vm does not want to properly pull from notabug and because I want to pull reliance off of dedicated servers and into major cdns.
    - doing regular git clone and do_vcs outside of script works but inside script says connection timed out with 0b transferred

I think that I might see if I can do automatic fallback for repos to <https://gitlab.com/media-autobuild_suite-dependencies> in do_vcs in case the original link stops failing for odd reasons.

(I really wish that github and other major git providers had a proper mirroring repos functionality [github needs access to the server's .git folder])

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -674,7 +674,7 @@ fi
 
 if [[ $ffmpeg != "no" ]] && enabled libsoxr; then
     _check=(soxr.h libsoxr.a)
-    if do_vcs https://notabug.org/RiCON/soxr.git libsoxr; then
+    if do_vcs "https://gitlab.com/media-autobuild_suite-dependencies/libsoxr.git"; then
         do_uninstall "${_check[@]}"
         do_cmakeinstall -DWITH_LSR_BINDINGS=off -DBUILD_TESTS=off -DWITH_OPENMP=off
         do_checkIfExist
@@ -1174,7 +1174,7 @@ fi
 
 _check=(DeckLinkAPI.h DeckLinkAPIVersion.h DeckLinkAPI_i.c)
 if [[ $ffmpeg != "no" ]] && enabled decklink &&
-    do_vcs "https://notabug.org/RiCON/decklink-headers.git"; then
+    do_vcs "https://gitlab.com/media-autobuild_suite-dependencies/decklink-headers.git"; then
     do_makeinstall PREFIX="$LOCALDESTDIR"
     do_checkIfExist
 fi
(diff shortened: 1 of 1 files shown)


build/media-suite_compile.sh
```diff
@@ -1174,7 +1174,7 @@ fi
 
 _check=(DeckLinkAPI.h DeckLinkAPIVersion.h DeckLinkAPI_i.c)
 if [[ $ffmpeg != "no" ]] && enabled decklink &&
-    do_vcs "https://notabug.org/RiCON/decklink-headers.git"; then
+    do_vcs "https://gitlab.com/media-autobuild_suite-dependencies/decklink-headers.git"; then
```
wiiaboo: If you give me push permissions to this one, it could be the official repo.

1480c1:
I tried to add RiCON, but you did not show up in the settings page. Maybe a
privacy setting?

build/media-suite_compile.sh
```diff
@@ -1228,7 +1228,7 @@ if [[ $x264 != no ]]; then
         unset_extra_script
         if [[ $standalone = y && $x264 =~ (full|fullv) ]]; then
             _check=("$LOCALDESTDIR"/opt/lightffmpeg/lib/pkgconfig/libav{codec,format}.pc)
-            do_vcs "https://git.ffmpeg.org/ffmpeg.git"
+            do_vcs "https://github.com/FFmpeg/FFmpeg.git"
```
wiiaboo: I'm pretty sure this mirror is not automatically synced. As far as the suite is concerned that might be good enough, but for people that need to be directly using the official repo it might be better to leave this as it was and maybe prioritize repo fallbacks feature for this one.

wiiaboo:
Maybe you need to add me to the project as a member, I dunno, haven't used gitlab a lot.

build/media-suite_compile.sh
1480c1: Sure, I can change it back to ffmpeg.org or change it to the gitlab since that is updated hourly

1480c1:
I think you should have access now, I just saw your request

build/media-suite_compile.sh
wiiaboo: If you feel like not doing the fallbacks soonish, I'm also personally fine with still changing this URL to the github mirror as I'm not building the suite as frequently as I used to (my main driver now is a Mac as well).

1480c1: I'll probably start of the fallback function after I get home so I can do that then. TBH, I used to have only windows computers (laptop, 3 desktops, and 2 server 2012 r2,) but now I only have one windows desktop with everything else Linux based

wiiaboo:
Hm, I'd rather split commits that are more easily reverted and searchable.

1480c1:
Sure

Issue #1172, opened 2019-04-21
Vulkan Headers : parsing .json file always fails


pat357:
When the suite is trying the get the l**ast-known-good version** from the Vulcan Headers, the parsing using **jq** seems always to fail. {**jq gives a compile error**)
See the relevant output from the log below:

```
mujs git .................................................. [Up-to-date]
jq: error: select/1 is not defined at <top-level>, line 1:
.repos[] | select(.name == "Vulkan-Headers").commit           
jq: 1 compile error
(23) Failed writing body

media-autobuild_suite_647 (64bit):   Running git clone for vulkan-headers in .  Running git clone for vulkan-headers in .
+ [[ git = \s\v\n ]]
+ git clone https://github.com/KhronosGroup/Vulkan-Headers.git vulkan-headers-git
Cloning into 'vulkan-headers-git'...
remote: Enumerating objects: 917, done.

+ set +x
+ [[ git = svn ]]
+ [[ git = hg ]]
+ [[ git = git ]]
+ [[ -n https://github.com/KhronosGroup/Vulkan-Headers.git ]]
+ git remote set-url origin https://github.com/KhronosGroup/Vulkan-Headers.git
+ git config remote.origin.fetch '+refs/heads/*:refs/remotes/origin/*'
+ [[ -f .git/refs/heads/ab-suite ]]
+ git branch -f --no-track ab-suite
+ git checkout ab-suite
Switched to branch 'ab-suite'
++ vcs_getlatesttag v1.1.106
++ local ref=v1.1.106
++ [[ -n git ]]
++ [[ git != git ]]
++ local tag
++ [[ v1.1.106 = \L\A\T\E\S\T ]]
++ [[ v1.1.106 = \G\R\E\A\T\E\S\T ]]
++ [[ v1.1.106 != \v\1\.\1\.\1\0\6 ]]
++ echo v1.1.106
+ git reset --hard v1.1.106
HEAD is now at 08cbb54 scripts: Fix generator on Windows with Python < 3.6
++ git rev-parse HEAD
+ oldHead=08cbb5458f692d4778806775f65eb3dc642ddbbf
+ set +x
┌ vulkan-headers git ................................ [Recently updated]

**Comment by 1480c1** [comment]
Can you delete jq-git if it exists in your build folder and rerun?

**Comment by pat357** [comment]
I have deleted the jq , vulkan-headers and vulkan folders from /build and forced recompile of those, but the same result 
```
[...]

PR #1184, opened 2019-04-30
Vulkan: Use update_deps.py to get correct version of headers.

### LINKED ISSUE #1179: Vulkan header fails because of '-Wno-stringop-truncation'
[logs.zip](https://0x0.st/zb4_.zip)

1480c1:
Closes #1179 (?)
@Selur and @Helenerineium  , can you confirm if this fixes the error?

I think there might be a problem with jq still because of #1179 and #1172 . Will need to investigate

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1846,39 +1846,32 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
         do_checkIfExist
     fi
 
-    if ! mpv_disabled vulkan; then
-        # Get the known_good json and get the commit version of the Vulkan-Headers.
-        # It is assumed that the version number has quotes around them (sed)
-        [[ "$jq" != y ]] && do_pacman_install jq
-        _Vulkan_Version=$(
-            ${curl_opts[@]} https://raw.githubusercontent.com/KhronosGroup/Vulkan-ValidationLayers/master/scripts/known_good.json |
-            jq -r '.repos[] | select(.name == "Vulkan-Headers").commit')
-        [[ "$jq" != y ]] && do_pacman_remove jq
-        _check=(vulkan/vulkan.h)
-        if do_vcs "https://github.com/KhronosGroup/Vulkan-Headers.git#tag=${_Vulkan_Version:-v1.1.106}" vulkan-headers; then
+    _check=(libvulkan.a vulkan.pc)
+    if ! mpv_disabled vulkan &&
+        do_vcs "https://github.com/KhronosGroup/Vulkan-Loader.git" vulkan-loader; then
+        _shinchiro_patches="https://raw.githubusercontent.com/shinchiro/mpv-winbuild-cmake/master/packages"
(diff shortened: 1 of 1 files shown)


wiiaboo:
Alternatively, you could do like mpv and run the python tool inside vulkan-loader and then compile/install vulkan-headers in it: https://github.com/mpv-player/mpv/commit/4e9dc60d9e7728cddfc901a8b0b7048eeebed5e6

wiiaboo:
I mistook shaderc with vulkan-loader, but the same tool is there as well: https://github.com/KhronosGroup/Vulkan-Loader/blob/master/scripts/update_deps.py

1480c1:
I might actually do that since it seems that that would be more reasonable and maintainable

1480c1:
Problem with using the python script, it uses cmake's default generator, which for some, happens to be Visual Studio.
I've tried to change the settings using the json, but I can't seem to get the CMAKE_INSTALL_PREFIX to be set to `LOCALDESTDIR` but it doesn't parse it like normal.

Guess I have to learn python.

wiiaboo:
https://github.com/KhronosGroup/Vulkan-Loader/blob/master/scripts/update_deps.py#L565-L571

Just use `--no-build` and move vulkan-header compilation/installatio to inside vulkan-loaders.
There's also msvc-specific options for the update_deps.py build part, so it's better to just not build with it.

1480c1:
Yea, that's what I'm thinking to do

build/media-suite_compile.sh
```diff
@@ -1846,39 +1846,32 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
-        _check=(vulkan/vulkan.h)
-        if do_vcs "https://github.com/KhronosGroup/Vulkan-Headers.git#tag=${_Vulkan_Version:-v1.1.106}" vulkan-headers; then
+    _check=(libvulkan.a vulkan.pc)
+    if ! mpv_disabled vulkan &&
+        do_vcs "https://github.com/KhronosGroup/Vulkan-Loader.git" vulkan-loader; then
+        _shinchiro_patches="https://raw.githubusercontent.com/shinchiro/mpv-winbuild-cmake/master/packages"
+        do_uninstall "${_check[@]}"
+        do_patch "$_shinchiro_patches/vulkan-0001-cross-compile-static-linking-hacks.patch"
+        create_build_dir
+        log dependencies ../scripts/update_deps.py --no-build
```
wiiaboo: Better explicitely run it with `/usr/bin/python3`.

wiiaboo: mingw python3/python2 comes first in path and might not work.

```suggestion
        log dependencies /mingw64/bin/python3 ../scripts/update_deps.py --no-build
```

1480c1: wait, it is better to use msys2 python or mingw?

wiiaboo: Msys2 one.

wiiaboo:
Just use msys2 python and you don't have to worry about bitness either.

PR #1189, opened 2019-05-05
extra_script: Add recompile flag. Closes #1186 

### LINKED ISSUE #1186: Request: recompile flag.
Hi,

After the insertion of `*_custom.sh` support, it will be interesting to support some _extended_ functions. Yes, I know, this suite isn't created for development, but I can assist.

Now we have the "file flag" `custom_updated` to force to recompile. And this is great, as we can enforce to recompile. However, in some cases it will be interesting to have too:
- `not_reconfigure_and_clean` flag: It will be a complement to `custom_updated` and the objective is to not "reconfigure" and not "clean" after the compilation. So we can call the suite another time and the target package (for example FFmpeg) don't be reconfigured, but it will be recompiled from the last state. Then it will be more simple to apply some new patch (or fix some problem) and recompile without waiting for the entire process.

What you think?

1480c1:
Closes #1186

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1216,13 +1216,15 @@ log() {
 
 create_build_dir() {
     local build_dir="build${1:+-$1}-$bits"
-    if [[ "$(basename "$(pwd)")" = "$build_dir" ]]; then
-        rm -rf ./*
-    elif [[ -d "$build_dir" ]] && ! rm -rf ./"$build_dir"; then
-        cd_safe "$build_dir" && rm -rf ./*
-    else
-        mkdir "$build_dir" && cd_safe "$build_dir"
+    if [[ ! -f "$(get_first_subdir)/do_not_clean" ]]; then
+        if [[ "$(basename "$(pwd)")" = "$build_dir" ]]; then
+            rm -rf ./* && cd_safe ".."
+        elif [[ -d "$build_dir" ]] && ! rm -rf ./"$build_dir"; then
+            cd_safe "$build_dir" && rm -rf ./* && cd_safe ".."
+        fi
     fi
+    [[ ! -d "$build_dir" ]] && mkdir "$build_dir"
+    cd_safe "$build_dir"
 }
 
 get_external_opts() {
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -282,7 +282,12 @@ do_vcs() {
         do_print_status "┌ ${vcsFolder} ${vcsType}" "$orange" "Newer dependencies"
     else
         do_print_status "${vcsFolder} ${vcsType}" "$green" "Up-to-date"
-        return 1
+        if [[ -f recompile ]]; then
+            do_print_status "┌ ${vcsFolder} ${vcsType}" "$orange" "Forcing recompile"
+            do_print_status "${bold}├${reset} Found recompile flag" "$orange" "Recompiling"
+        else
+            return 1
+        fi
     fi
     return 0
 }
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -260,7 +260,7 @@ do_vcs() {
     rm -f custom_updated
     check_custom_patches
 
-    if { [[ "$oldHead" != "$newHead" ]] || [[ -f custom_updated ]]; }; then
+    if [[ "$oldHead" != "$newHead" ]] || [[ -f custom_updated ]]; then
         touch recently_updated
         rm -f ./build_successful{32,64}bit{,_*}
         if [[ $build32 = "yes" && $build64 = "yes" ]] && [[ $bits = "64bit" ]]; then
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1086,6 +1086,8 @@ do_cmake() {
     create_build_dir
     [[ $1 && -d "../$1" ]] && root="../$1" && shift
     extra_script pre cmake
+    [[ -f "$(get_first_subdir)/do_not_reconfigure" ]] &&
+        return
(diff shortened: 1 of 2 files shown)


lars18th:
Hi @1480c1 ,

This change does the same as `custom_updated` flag as I see.

The request has this logic:
- If the file "not_reconfigure_and_clean" exists then:
 -- If "not configured" then CONFIGURE
   ----  else no_configure (pass it).
 -- MAKE
 -- no_clean (pass it).
- else
 -- CONFIGURE
 -- MAKE
 -- CLEAN


Regards.

1480c1:
I will add it to do_configure and do_cmake later. I was busy studying and doing my exams so I have been unable to really focus on this

1480c1:
Should this skip pre and post configure section of the extra file?

lars18th:
> I will add it to do_configure and do_cmake later. I was busy studying and doing my exams so I have been unable to really focus on this

OK. Don't worry then!

> Should this skip pre and post configure section of the extra file?

I'll re-check it.

lars18th:
Hi @1480c1 ,

When creating the file `/build/ffmpeg-git/not_reconfigure_and_clean` and running the suite... (with this patch applyed):

The first time it configures and compiles. But:
1. After ending the compilation, the directory `/build/ffmpeg-git/build-static-64bit` is removed. Remember: the idea is to **not** remove the compilation to restart it after doing some changes in patches.
2. Also if you interrupt the compilation in the "make" section (after the configure), and then retart the suite, the compilation of FFmpeg fails... the "configure" is reused, but the compilation fails.

Please, try to resolve these problems.
Regards.

1480c1:
I still haven't done ffmpeg yet, I don't have much free time atm

lars18th:
OK. So still waiting.

This is only in response to...

>>Should this skip pre and post configure section of the extra file?

>I'll re-check it.

Regards.

1480c1:
I mean should I make it so that it ships pre and post configure

1480c1:
Right now it's skipping any pre and post configure sections

1480c1:
I still need to add documentation to readme

lars18th:
OK. I'll test it.

wiiaboo:
Btw, this will/should only work for out-of-root builds. The others require the clean or else broken builds will occur if multi-arch builds are being run.

1480c1:
Should I make it so that if the current directory matches get_first_subdir, then clean else skip?

wiiaboo:
Not sure. All the times a `make clean` is actually done manually is for in-root builds, so for out-of-root, all this needs to do is not delete the build directories before creating them.

1480c1:
Okay, then should I just remove the 
``` bash
   if [[ $_cmd =~ ^(make|ninja)$ ]]; then
        if [[ ${@: -1} =~ ^(distclean|clean)$ ]] &&
            [[ ! -f "$(get_first_subdir)/do_not_clean" ]] &&
            [[ "$PWD" != "$(get_first_subdir)" ]]; then
            echo "Skipping clean"
            return
        fi
```
?

wiiaboo:
Sure. The biggest usecase for this is probably ffmpeg/mpv, I'd guess.

mpv does build out-of-root, but we're not using `--out build-$bits`, so it's building always to the same `build` directory for both bitness.

wiiaboo:
mpv's distclean can be replaced with a basic `rm -rf ./build-$bits`, because that's what waf does.

1480c1:
I see that most of the calls to log make distclean and clean are in root, except xavs2, davs2, mediainfo

wiiaboo:
The remaining distclean in builds like FFmpeg are to cleanup older in-root builds because configures usually check if the root already contains config artifacts and prevents out-of-root builds.

Those three should be safe to ignore, unless explicitely requested.

Mediainfo's cmake was broken last time I tried it, and out-of-root builds with autoconf are also broken.

build/media-suite_helper.sh
```diff
@@ -1216,13 +1216,14 @@ log() {
...
-    else
-        mkdir "$build_dir" && cd_safe "$build_dir"
+    if [[ ! -f "$(get_first_subdir)/do_not_clean" ]]; then
+        if [[ "$(basename "$(pwd)")" = "$build_dir" ]]; then
+            rm -rf ./*
+        elif [[ -d "$build_dir" ]] && ! rm -rf ./"$build_dir"; then
+            cd_safe "$build_dir" && rm -rf ./* && cd_safe ".."
+        fi
     fi
+    mkdir "$build_dir" 2>/dev/null && cd_safe "$build_dir"
```
wiiaboo: `mkdir` does not return zero if the directory it's trying to create already exists, so the cd_safe won't run every time.

build/media-suite_helper.sh
```diff
@@ -1216,13 +1216,14 @@ log() {
...
     local build_dir="build${1:+-$1}-$bits"
-    if [[ "$(basename "$(pwd)")" = "$build_dir" ]]; then
-        rm -rf ./*
-    elif [[ -d "$build_dir" ]] && ! rm -rf ./"$build_dir"; then
-        cd_safe "$build_dir" && rm -rf ./*
-    else
-        mkdir "$build_dir" && cd_safe "$build_dir"
+    if [[ ! -f "$(get_first_subdir)/do_not_clean" ]]; then
+        if [[ "$(basename "$(pwd)")" = "$build_dir" ]]; then
+            rm -rf ./*
```
wiiaboo: You should do like below and `cd_safe ..`, otherwise you'll be creating another build_dir inside build_dir.

build/media-suite_helper.sh
1480c1: Fixed

build/media-suite_helper.sh
1480c1: fixed

PR #1190, opened 2019-05-05
zip_logs: Add no_logs flag to disable zipping and uploading logs

### LINKED ISSUE #1182: Disable logs uploads
Hi,

Any option to disable the logs upload functionality?

I'm using the `*_extra.sh` file to apply custom patches, and sometimes any of my patches fails to complie. So in this case it's irrelevant to upload the log.

Ideas?

1480c1:
Closes #1182 

To disable uploading logs to 0x0, add `touch "$LOCALBUILDDIR/no_logs"` to *_extra.sh

This disables uploading logs globally regardless of the package failing to compile.
(@wiiaboo should this be changed to upload if the local package has no modification?)

There is nothing to remove the flag so if log uploading is desired, remove the flag by hand.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -71,7 +71,8 @@ if [[ $noMintty = y ]]; then ( set -o posix ; set )>"$LOCALBUILDDIR/old.var"; fi
 
 source "$LOCALBUILDDIR"/media-suite_helper.sh
 
-echo -e "${orange}Warning: We will not accept any issues lacking any form of logs or logs.zip! ${reset}"
+[[ -f "$LOCALBUILDDIR/no_logs" ||  $logging = n ]]
+    && echo -e "${orange}Warning: We will not accept any issues lacking any form of logs or logs.zip!${reset}"
 
 buildProcess() {
 set_title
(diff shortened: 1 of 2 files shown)


wiiaboo:
Add a big fat warning to the build script start that we won't accept any issues lacking logs.

I'm fine with not uploading the logs if the file is still there. It's the user's responsability to keep it or delete it.

build/media-suite_compile.sh
```diff
@@ -71,6 +71,8 @@ if [[ $noMintty = y ]]; then ( set -o posix ; set )>"$LOCALBUILDDIR/old.var"; fi
 
 source "$LOCALBUILDDIR"/media-suite_helper.sh
 
+echo -e "${orange}Warning: We will not accept any issues lacking any form of logs or logs.zip! ${reset}"
```
wiiaboo: You can put it behind the same check in the previous commit as well.

wiiaboo:
```suggestion
[[ -f "$LOCALBUILDDIR/no_logs" ||  $logging = n ]]
    && echo -e "${orange}Warning: We will not accept any issues lacking any form of logs or logs.zip!${reset}"
```

lars18th:
LGTM

PR #1198, opened 2019-05-08
compile: fix regression caused by #1190. Fixes #1193

### LINKED ISSUE #1193: Just updated the suite after the latest commits and after the update check it just closes
Just updated the suite after the latest commits thats been oushed and after the update check it just closes. Not logs etc, it's just closes.

1480c1:
I am assuming that this fixes #1193 as when I switched noMintty off, the mintty terminal with compile.sh opened, but then immediately closed. I don't know why it worked with noMintty on.


wiiaboo:
Yeah, I can reproduce that noMintty issue with mujs.

1480c1:
The problem goes away when using a separated env using `env -i PATH=$PATH $(which make)`, which I guess means that something within the variables being set is causing an issue.

PR #1206, opened 2019-05-11
bat: separate all variables

1480c1:


1480c1:
With this, we can revert 8566e2514ca7befcbc8b379fbb9cf5efac4ff8c7 or keep it. I still can't figure out what is causing the problem

1480c1:
Although it may be better to leave it as it is since it would be safer that way.

PR #1217, opened 2019-05-29
gnutls: update to 3.6.8

schmidthubert:


wiiaboo:
Is git still broken?

schmidthubert:
Never was for me! But ./bootstrap still take its time!

wiiaboo:
Ah right, that was the reason for these. Is there no way to automate these? Maybe using the .sig files for integrity checking instead of hashes?

schmidthubert:
This is all for you to to decide! For me personally a minute more or less doesn't matter, I always have enough CPU power available ...

1480c1:
I could try to add a way to do_wget to accept a link to a sig file for the -h option if that would be desirable, although I do need to make sure how to confirm integrity with a sig file.

I'm not sure if it would be possible to automate looking for sig files since I don't think that  all of the repos that use do_wget use the same filestructure for archive and sig

wiiaboo:
We need to:
1. get the packager(s)' signatures which will be saved hardcoded in the suite (not dynamically retrieved as that's not as safe)
2. add them to the local gpg store if needed, and local-trust them
3. retrieve the .sig file, which is usually present in the same directory and with the same name as the file plus `.sig`
4. run `gpg --verify file.tar.gz.sig` in the same directory as `file.tar.gz.sig`

It's a bit more complex than just checking for hashes, but it's mostly safer to tamper.

This is how arch/pacman does sig checking:
ex pkgbuild: https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/gnutls#n18
https://git.archlinux.org/pacman.git/tree/scripts/libmakepkg/integrity/verify_signature.sh.in#n119

1480c1:
How should the key be coded in?

1480c1:
Do we want to have them alongside the url information etc? or have them in a separate location?

wiiaboo:
For now this can be merged.

PR #1219, opened 2019-05-30
Create cmake toolchain file and use it to set some sane values

wiiaboo:
Any way to make this a little better, @1480c1 ? I think you're way more comfortable with cmake than I am.

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1747,12 +1747,10 @@ create_cmake_toolchain() {
     local _win_path_LOCALDESTDIR="$(cygpath -m $LOCALDESTDIR)"
     local _win_path_MINGW_PREFIX="$(cygpath -m $MINGW_PREFIX)"
     local toolchain_file=(
-        "SET(CMAKE_C_COMPILER gcc)"
-        "SET(CMAKE_CXX_COMPILER g++)"
         "SET(CMAKE_RC_COMPILER_INIT windres)"
         "SET(PKG_CONFIG_EXECUTABLE $_win_path_LOCALDESTDIR/bin/ab-pkg-config-static.bat)"
         ""
-        "LIST(APPEND CMAKE_PROGRAM_PATH $_win_path_LOCALDESTDIR/bin ...)"
+        "LIST(APPEND CMAKE_PROGRAM_PATH $_win_path_LOCALDESTDIR/bin)"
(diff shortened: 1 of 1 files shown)


1480c1:
I can try, although I may need a bit of time to see exactly what is needed here

wiiaboo:
It did fix the libtiff "issue" with Cmake trying to find msys2 libm, at least. Didn't test with any other cmake lib to see if anything broke.

build/media-suite_helper.sh
```diff
@@ -1743,6 +1743,27 @@ EOF
             cat > "$LOCALDESTDIR"/bin/ab-pkg-config-static.bat
 }
 
+create_cmake_toolchain() {
+    local _win_path_LOCALDESTDIR="$(cygpath -m $LOCALDESTDIR)"
+    local _win_path_MINGW_PREFIX="$(cygpath -m $MINGW_PREFIX)"
+    local toolchain_file=(
+        "SET(CMAKE_C_COMPILER gcc)"
```
1480c1: Is it necessary to hardcode gcc? Maybe better to let cmake detect c compiler/use CC variable

wiiaboo: Sure, those three lines were sort of copied from https://github.com/shinchiro/mpv-winbuild-cmake/blob/master/toolchain.cmake.in

1480c1: I see, I see

build/media-suite_helper.sh
```diff
@@ -1743,6 +1743,25 @@ EOF
...
 }
 
+create_cmake_toolchain() {
+    local _win_path_LOCALDESTDIR="$(cygpath -m $LOCALDESTDIR)"
+    local _win_path_MINGW_PREFIX="$(cygpath -m $MINGW_PREFIX)"
+    local toolchain_file=(
+        "SET(CMAKE_RC_COMPILER_INIT windres)"
+        "SET(PKG_CONFIG_EXECUTABLE $_win_path_LOCALDESTDIR/bin/ab-pkg-config-static.bat)"
+        ""
+        "LIST(APPEND CMAKE_PROGRAM_PATH $_win_path_LOCALDESTDIR/bin ...)"
```
1480c1: What's with the triple dot?

wiiaboo: Probably shitty copy/paste from docs and thinking they do something like in JS.

1480c1:
Other than the potential problem of quotes around the paths, I don't see anything wrong/needing to change

wiiaboo:
Real big brain stuff, didn't realize I'd pushed the commit to master already.

1480c1:
Lul

PR #1221, opened 2019-05-31
helper: add check_file_integrity

1480c1:
```
We need to:
1. get the packager(s)' signatures which will be saved hardcoded in the suite (not dynamically retrieved as that's not as safe)
2. add them to the local gpg store if needed, and local-trust them
3. retrieve the .sig file, which is usually present in the same directory and with the same name as the file plus `.sig(n)?|asc`
4. run `gpg --verify file.tar.gz.sig` in the same directory as `file.tar.gz.sig`
```
This is how arch/pacman does sig checking:
ex pkgbuild: https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/gnutls#n18
https://git.archlinux.org/pacman.git/tree/scripts/libmakepkg/integrity/verify_signature.sh.in#n119

_Originally posted by @wiiaboo in https://github.com/jb-alvarado/media-autobuild_suite/pull/1217#issuecomment-497252379_

Going to move the conversation here. Still a WIP,

How should we store certain information such as the keys etc? 
I'm thinking of either modifying do_wget with another flag to accept a url to a sig file or 

I need to go now, I will edit appropriately

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ update_suite.sh
 !build/patches/
 !build/extras/
 /build/post_suite.sh
+!build/files.json
 
 ## VS Code
 .vscode
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ For information about the compiler environment see the wiki, there you also have
     - ffmbc (git) (unsupported)
     - flac (git)
     - haisrt tools (git)
+    - jo (git)
     - jq (git)
     - kvazaar (git)
     - lame (3.100)
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ For information about the compiler environment see the wiki, there you also have
             - enabled by default, use --disable-sdl2 if unneeded
     - Zeranoe-emulating build (in addition to Light)
         - avisynth (needs avisynth dll installed)
-        - fontconfig (2.13.1)
+        - fontconfig (latest release)
         - only one of these TLS libs (including schannel) can be enabled at once:
             - openssl (mingw)
                 - preferred to gnutls and to libtls if all three are in options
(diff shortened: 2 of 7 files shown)


wiiaboo:
Maybe something like this? https://jqplay.org/s/9a_OBo6FlE

1480c1:
That seems pretty good, although there is still the question of are we going to put it in the script or host it on GitHub.io?

1480c1:
That seems pretty good, although there is still the question of are we going to put it in the script or host it on GitHub.io?

RiCONfru:
Seems fine to leave it in the repo.

wiiaboo:
Seems fine to leave it in the repo as an actual .json that gets updated along with the suite.

wiiaboo:
Maybe we could leave the possibility for easier change of the actual URL/repo for those people in countries/networks that block certain domains.

Like a files.custom.json where people would just need to change whichever URLs they need. This should of course not be officially supported since people are then responsible for keeping track of updates to URLs/repos if they choose to override the default ones.

wiiaboo:
But yeah, so far, looks good.

wiiaboo:
Interesting, I'd never seen how a fixup to a fixup looks.

1480c1:
Doing `git commit --fixup=<commit ID>` prepends `fixup!`, after that, do `git rebase upstream/master --autosquash -i` and it will mark those commits as fixup automatically

ImportTaste:
Is this ready for testing? If so, what is the recommended way to verify it is detecting an invalid file?

1480c1:
Not yet, still haven't actually added the logic completely, trying to clean up other subprojects

PR #1223, opened 2019-05-31
upx: update to 3.95

1480c1:
Do we need to worry about a 64bit version of upx?

I'm not entirely sure about this logic down below. It should be
``` bash
if [[ $build64 = y ]]; then
    do_wget -h 5c076f87ba64d82f11513f4af0ceb07246a3540aa3c72ca3ffc2d53971fa56e3 \
        "https://github.com/upx/upx/releases/download/v3.95/upx-3.95-win64.zip"
else
    do_wget -h f94ff30b175d125d1c238458716f5808aee222547a813918b44d0f67035c0054 \
        "https://github.com/upx/upx/releases/download/v3.95/upx-3.95-win32.zip"
fi
```
(haven't used brackets before really.)

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -133,9 +133,9 @@ set_title "compiling global tools"
 echo -e "\n\t${orange}Starting $bits compilation of global tools${reset}"
 
 if [[ $packing = y ]] &&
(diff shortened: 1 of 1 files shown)


wiiaboo:
Nah, I think 32-bit is enough, at least until Windows copies Mac in not allowing 32-bit binaries anywhere.

1480c1:
Okay, only doing 32-bit  for upx

PR #1224, opened 2019-05-31
faac: change to github from sf

1480c1:
One less do_wget we would need to worry about

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -605,8 +605,8 @@ fi
 [[ $faac = y ]] && do_pacman_install faac
 _check=(bin-audio/faac.exe)
 if [[ $standalone = y && $faac = y ]] && ! files_exist "${_check[@]}" &&
-    do_wget_sf -h 2b58d621fad8fda879f07b7cad8bfe10 \
-        "faac/faac-src/faac-1.29/faac-1.29.9.2.tar.gz"; then
+    do_wget -h e23a05f13f9695a81f250c30bd30e5bd636a0f6891a9ea9093ce4bfbf758217b \
+        "https://github.com/knik0/faac/archive/1_29_9_2.tar.gz"; then
     do_uninstall libfaac.a faac{,cfg}.h "${_check[@]}"
     [[ $standalone = y ]] || sed -i 's|frontend||' Makefile.am
     do_separate_conf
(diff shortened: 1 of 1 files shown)


1480c1:
Actually, which is better, using do_vcs with branch to the tag, or using the releases?

I personally prefer using do_vcs so that the extra_*.sh can be used, but I may just as well add in the extra_*.sh to do_wget if it detects a repo

wiiaboo:
You could add a check for do_pkgConfig or checking the version in the produced binary if you want to avoid running the do_vcs logic.

I don't know how many people still compile this useless encoder when they could just be using fdk-aac instead. Maybe we should start doing some telemetry on how many compiles people do and what options they use and we start dropping useless libs.

schmidthubert:
Sorry to disagree here. Support of anything should be dropped by ffmpeg team first. Telemetry, surveys etc are only snapshots in time. Like electing Trump or voting for or against Brexit it could all look different the next day. Most of the libs which are not developed any further or only very little don't need much "support" from your end and simply compile just fine. BTW: (Although I don't like Python stuff ...) I keep an eye on [DeadSix27's script](https://github.com/DeadSix27/python_cross_compile_script) as well because he has most supported libs including most patches needed. So in case of problems you can have a peek there as well :+1:

1480c1:
Changed to just use the release from github

1480c1:
I will try to implement the extra_*.sh for do_wget soon, potentially alongside the integrity check

build/media-suite_compile.sh
```diff
@@ -605,8 +605,8 @@ fi
 [[ $faac = y ]] && do_pacman_install faac
 _check=(bin-audio/faac.exe)
 if [[ $standalone = y && $faac = y ]] && ! files_exist "${_check[@]}" &&
-    do_wget_sf -h 2b58d621fad8fda879f07b7cad8bfe10 \
-        "faac/faac-src/faac-1.29/faac-1.29.9.2.tar.gz"; then
+    do_wget -h e23a05f13f9695a81f250c30bd30e5bd636a0f6891a9ea9093ce4bfbf758217b \
+        "https://github.com/knik0/faac/archive/1_29_9_2.tar.gz"; then
```
wiiaboo: These git archive tarballs usually need a specific dirname. In this case it gets extracted to "faac-1_29_9_2", so you need to take this into account:
```suggestion
        "https://github.com/knik0/faac/archive/1_29_9_2.tar.gz" "faac-1_29_9_2.tar.gz"; then
```
or
```suggestion
        "https://github.com/knik0/faac/archive/1_29_9_2.tar.gz" "" "faac-1_29_9_2"; then
```

wiiaboo: They use the format "${reponame}-${branch || tag}.tar.gz"

wiiaboo:
I'm not saying we should drop stuff no one is using, but if it starts creating too many issues and no one cares for it, there's no point trying to hack away a fix when it could just be removed.

PR #1225, opened 2019-05-31
vswhere: use latests

1480c1:
https://help.github.com/en/articles/linking-to-releases
Since I found out we can link to the latest release.
Although this won't work for repos that hardcodes the version in the filename


wiiaboo:
I forgot about that functionality.

PR #1229, opened 2019-06-02
Github: Use markdown issue templates

1480c1:
Idk about the pr templates since I can't create a pr from a branch on a fork to the master of a fork, but the issue templates seem to work. The PR templates require the files to be in the master branch of the master repo. I can split the commits into individual templates for easy reverting in wanted.

Motivated by the recent lack of issue description in the title and lack of information of 32|64 bit builds etc.

e.g. "Compilation failed", "x265", "libwebp", etc

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,6 +1,0 @@
-If you use the built-in options and it's broken because of a lib you didn't even
-know existed or will never use, please choose your own libs and avoid creating
-issues about it. I barely have any time or will to maintain this.
-
-For new libs and tools pull requests are strongly recommended instead of opening
-a feature request for the same reason above.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,25 @@
+---
+name: 'Build Bug report'
+about: 'Repo bug report'
+title: '[mingw32|mingw64] [repo]'
+labels: 'Bug Report'
+assignees: ''
+
+---
+<!---
+If you use the built-in options and it's broken because of a lib you didn't even know existed or will never use, please choose your own libs and avoid creating issues about it. I barely have any time or will to maintain this.
+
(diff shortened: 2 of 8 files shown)


1480c1:
it seems that there can only be one pr template.

1480c1:
https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository#adding-a-pull-request-template

wiiaboo:
Interesting, how do those new/update PR templates look like?

1480c1:
Sadly, the pr templates require you to modify the link to use it
something like: `https://github.com/jb-alvarado/media-autobuild_suite/compare/master...1480c1:issuetemp?expand=1&template=new.md`
the `&template=new.md` would pull the new.md template

1480c1:
I might remove the 
```
---
name: 'New Package'
about: 'Add a new package'
title: '[repo]: Add [repo]'
labels: 'New Package'
assignees: 'wiiaboo'

---
```
since it does not seem that those get applied properly for pr templates, only issue templates.

1480c1:
Seems that those templates are mainly for automation etc with curl etc

wiiaboo:
Can you check if the following ideas are feasible (no priority):
1. Clearing the compilation window if logging=y, to prevent people from copying mostly useless trash text. Just a few lines with the link to search issues, so people can check if there are already open issues for this specific failure with:
   - `$(get_first_subdir)`
   - maybe the step where the build failed (configure/build/install)
   - maybe the short commit ID
2. Making the upload of logs a question, where if you [Enter], they're uploaded and you get only simple pastable text that fits in the terminal, otherwise they can just close the window and nothing is uploaded but the logs.zip is still created before the question. The suite option to never upload logs can still be respected, so in case that's disabled, the question is never made.
3. Creating templates where either the link points to the creation of a new issue, which the user then clicks after copying the logs.zip markdown text with the link, with:
   - title of the issue is pre-filled with something like '_libsomething or $(get_first_subdir)_: _configure_ failed with commit _short-commit-id_' where the italics are easily retrievable variables by the suite.

1480c1:
When I get my hotel, I will try and see

PR #1230, opened 2019-06-03
ffmbc: fix build

schmidthubert:
Fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/1202

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2140,10 +2140,7 @@ fi
 _check=(bin-video/ffmbc.exe)
 if [[ $ffmbc = y ]] && do_vcs "https://github.com/bcoudurier/FFmbc.git#branch=ffmbc"; then # no other branch
     _notrequired=yes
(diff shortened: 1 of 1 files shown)


schmidthubert:
Also fixes https://github.com/jb-alvarado/media-autobuild_suite/issues/959

wiiaboo:
Since it's a do_vcs build, you could use a git patch instead and try to submit a PR to upstream and apply it here meanwhile.
It's also more readable than a bunch of seds.

schmidthubert:
I prefer not to PR to upstream directly in this case because ffmbc is more than behind to say the least and it seems to use a lot of (other) files (as well) from ffmpeg directly. Since os_support and many other files from ffmpeg have already been fixed (for ages) and the author has been informed as well (but apparently refuses to reply for whatever reason) I think a patch should do it for now. Unfortunately issues are also disabled for that repo( hint hint ... project is dead?), I'd have created one otherwise with a link to the patch, but not so. Anyway, tested the patch and it's working OK.

BlackScreen:
https://github.com/bcoudurier/FFmbc/commit/a751c9271aa2514e960640c4523f8b39b4637068
It seems, it is not completely dead...

PR #1236, opened 2019-06-05
VapourSynth: static linking

Helenerineium:
This patch makes ffmpeg and mpv no longer need extra dlls at startup when compiling with vapoursynth.

The reason why I am not using gendef and dlltool for vapoursynth:

> One distinction is that when I'd tested with using gendef to handle VapourSynth too, I could get FFmpeg to link to it, but mpv refused to link to libavformat. gendeffing only Python and then cross-compiling VapourSynth like any other library results in both FFmpeg and mpv being able to link to it (importantly, mpv being able to directly play back scripts this way).

See [this post](http://forum.doom9.org/showthread.php?p=1875818#post1875818) on Doom9.

Do we need to checkout release tags or master will be OK?

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1718,6 +1718,8 @@ if [[ $ffmpeg != "no" ]]; then
             do_patch "https://0x0.st/zeB6.txt"
         fi
 
+        enabled vapoursynth && do_patch "https://github.com/Helenerineium/FFmpeg/commit/09af1ed650cfd221282ca47b851ad96a4bfcc700.patch" am
+
         # shared
         if [[ $ffmpeg != "static" ]] && [[ ! -f build_successful${bits}_shared ]]; then
             do_print_progress "Compiling ${bold}shared${reset} FFmpeg"
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1497,53 +1497,67 @@ if enabled libsrt && do_vcs "https://github.com/Haivision/srt.git"; then
 fi
 
 
-if get_vs_prefix &&
-    { ! mpv_disabled vapoursynth || enabled vapoursynth; }; then
-    vsversion="$("$vsprefix"/vspipe.exe -v | grep -Po "(?<=Core R)\d+")"
-    _check=(lib{vapoursynth,vsscript}.a vapoursynth{,-script}.pc
-        vapoursynth/{VS{Helper,Script},VapourSynth}.h)
-    if ! pc_exists "vapoursynth = $vsversion" || ! files_exist "${_check[@]}"; then
-        do_vcs "https://github.com/vapoursynth/vapoursynth.git"
-        do_uninstall {vapoursynth,vsscript}.lib "${_check[@]}"
-        if git show-ref -q "R${vsversion}"; then
-            git reset -q --hard "R${vsversion}"
-        else
-            git reset -q --hard origin/master
-        fi
+if  { ! mpv_disabled vapoursynth || enabled vapoursynth; }; then
+    _python_ver=3.7.3
+    _python_lib=python37
+    [[ $bits = 64bit ]] && _arch=amd64 || _arch=win32
+    _check=("lib$_python_lib.a")
+    if files_exist "${_check[@]}"; then
+        do_print_status "python $_python_ver" "$green" "Up-to-date"
+    elif do_wget "https://www.python.org/ftp/python/$_python_ver/python-$_python_ver-embed-$_arch.zip"; then
+        gendef "$_python_lib.dll" >/dev/null 2>&1
+        dlltool -y "lib$_python_lib.a" -d "$_python_lib.def"
+        [[ -f "lib$_python_lib.a" ]] && do_install "lib$_python_lib.a"
+        do_checkIfExist
+    fi
 
-        do_install include/*.h include/vapoursynth/
+    _vsver=46
+    _check=(lib{vapoursynth,vsscript}.a vapoursynth{,-script}.pc vapoursynth/{VS{Helper,Script},VapourSynth}.h)
+    if pc_exists "vapoursynth = $_vsver" && files_exist "${_check[@]}"; then
+        do_print_status "vapoursynth R$_vsver" "$green" "Up-to-date"
+    elif do_wget "https://github.com/vapoursynth/vapoursynth/releases/download/R$_vsver/VapourSynth${bits%bit}-Portable-R$_vsver.7z"; then
+        do_uninstall {vapoursynth,vsscript}.lib "${_check[@]}"
+        do_install sdk/include/*.h include/vapoursynth/
 
         create_build_dir
         for _file in vapoursynth vsscript; do
-            gendef - "$vsprefix/${_file}.dll" 2>/dev/null |
-                sed -r -e 's|^_||' -e 's|@[1-9]+$||' > "${_file}.def"
-            dlltool -l "lib${_file}.a" -d "${_file}.def" \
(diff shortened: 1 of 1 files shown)


wiiaboo:
Last time I tried this, mpv now depended on mingw's python's dlls; does this not happen anymore?

wiiaboo:
Also, does static vapoursynth work with compiled plugins? If not, there's little point in it.

Helenerineium:
> Last time I tried this, mpv now depended on mingw's python's dlls; does this not happen anymore?

> Also, does static vapoursynth work with compiled plugins? If not, there's little point in it.

Yes and yes, at least it works on my side.
May need to wait for other people to chime in.

wiiaboo:
And whether release or master, depends on how stable master is. If it breaks easily or takes too long to generate autoconf stuff, releases are better.

schmidthubert:
I have some questions too before I even bother to my waste my time: trying this: 
1) why was the check for the installed VapourSynth version removed?
2) will an *installed* *SAY* VS R45 (not R45.1!) work with *installed* python 3.7 (on the Windows machine!) even if mingw/msys comes with 3.7.3 already?

I won't even bother visiting VS or Python sites here to look up what works with what and what doesn't.

wiiaboo:
If you don't know how to test a vapoursynth script, you're just basically testing compilation and execution without actually using VS. That's fine, but not enough for this to be approved. Most people don't use just plain built-in VS in their scripts, I assume.

Helenerineium:
> If you don't know how to test a vapoursynth script, you're just basically testing compilation and execution without actually using VS. That's fine, but not enough for this to be approved. Most people don't use just plain built-in VS in their scripts, I assume.

`ffmpeg -f vapoursynth -i test.vpy
`

````
from vapoursynth import core
clip = core.ffms2.Source(r'test.mp4')
clip = clip.std.BoxBlur(hradius=2, vradius=2).misc.AverageFrames(weights=[1] * 5)
clip.set_output()
````

**Comment by wiiaboo** [comment]
Hey, nice, it does work. Both with FFmpeg and mpv.
Is this ready for merge then, or are you thinking of doing something more?

**Comment by wiiaboo** [comment]
Though it does depend on Python being installed (or embedded), as expected, it also still does depend on VSScript.dll, though?

**Comment by Helenerineium** [comment]
> Though it does depend on Python being installed (or embedded), as expected, it also still does depend on VSScript.dll, though?

This depends on how you define "depend".

Previously, ffmpeg / mpv compiled with vapoursynth support will need VSScript.dll to run even if you don't use the vapoursynth demuxer (e.g. a simple `-version` command)

Now vapoursynth and python (installed or embedded) is only needed if you want to use the vapoursynth demuxer.

> Is this ready for merge then, or are you thinking of doing something more?

I plan to use another vapoursynth demuxer implementation found [here](https://pastebin.com/Ls1LmTyj).

> [It seems that there is something wrong with wm4's vapoursynth demuxer implementation.](http://forum.doom9.org/showthread.php?p=1867074#post1867074)

**Comment by wiiaboo** [comment]
I thought since we're compiling vapoursynth statically vsscript.dll would no longer be a dependency, runtime or delay-import. Why not just gendef a delay import lib for vapoursynth then?

**Comment by Helenerineium** [comment]
Since R46 is released, the portable package now includes the necessary header files, pkgconfig files still need to be manually created, though.

I will be pushing more changes this weekend.

File: build/media-suite_compile.sh [hunk 1]
```diff
@@ -1497,53 +1497,74 @@ if enabled libsrt && do_vcs "https://github.com/Haivision/srt.git"; then
...
-        if git show-ref -q "R${vsversion}"; then
-            git reset -q --hard "R${vsversion}"
-        else
-            git reset -q --hard origin/master
-        fi
+if  { ! mpv_disabled vapoursynth || enabled vapoursynth; }; then
+
+    _ver=3.7.3
+    _file=python37
+    [[ $bits = 64bit ]] && arch=amd64 || arch=win32
```
wiiaboo: Either unset `$arch` somewhere after you don't need it or use `$_arch` instead, this variable is used by FFmpeg with different values.

build/media-suite_compile.sh
```diff
@@ -1497,53 +1497,74 @@ if enabled libsrt && do_vcs "https://github.com/Haivision/srt.git"; then
...
+        for _files in vapoursynth vsscript; do
+            gendef "../$_files.dll" >/dev/null 2>&1
+            dlltool -y "lib$_files.a" -d "$_files.def"
+            [[ -f lib$_files.a ]] && do_install "lib$_files.a"
         done
 
+        printf "%s\n" "prefix=$LOCALDESTDIR" > vapoursynth.pc
+        printf "%s\n" "prefix=$LOCALDESTDIR" > vapoursynth-script.pc
+
+        sed -i \
```
wiiaboo: Why not just use printf all the way?

build/media-suite_compile.sh
```diff
@@ -1497,53 +1497,74 @@ if enabled libsrt && do_vcs "https://github.com/Haivision/srt.git"; then
...
-        do_uninstall {vapoursynth,vsscript}.lib "${_check[@]}"
-        if git show-ref -q "R${vsversion}"; then
-            git reset -q --hard "R${vsversion}"
-        else
-            git reset -q --hard origin/master
-        fi
+if  { ! mpv_disabled vapoursynth || enabled vapoursynth; }; then
+
+    _ver=3.7.3
+    _file=python37
```
wiiaboo: This variable name doesn't really make sense.
Use `$_python_ver` and `$_python_libname` instead.

build/media-suite_compile.sh
```diff
@@ -1497,53 +1497,74 @@ if enabled libsrt && do_vcs "https://github.com/Haivision/srt.git"; then
...
         done
 
+        printf "%s\n" "prefix=$LOCALDESTDIR" > vapoursynth.pc
+        printf "%s\n" "prefix=$LOCALDESTDIR" > vapoursynth-script.pc
+
+        sed -i \
+            -e 'a exec_prefix=${prefix}' \
+            -e 'a libdir=${exec_prefix}/lib' \
+            -e 'a includedir=${prefix}/include/vapoursynth' \
+            -e 'a\ ' \
```
wiiaboo: These blanks lines are not needed.

build/media-suite_compile.sh
```diff
@@ -1497,53 +1497,74 @@ if enabled libsrt && do_vcs "https://github.com/Haivision/srt.git"; then
...
+            [[ -f lib$_files.a ]] && do_install "lib$_files.a"
         done
 
+        printf "%s\n" "prefix=$LOCALDESTDIR" > vapoursynth.pc
+        printf "%s\n" "prefix=$LOCALDESTDIR" > vapoursynth-script.pc
+
+        sed -i \
+            -e 'a exec_prefix=${prefix}' \
+            -e 'a libdir=${exec_prefix}/lib' \
+            -e 'a includedir=${prefix}/include/vapoursynth' \
```
wiiaboo: These three lines are the same for both files. It's easier to make sure they are the same if you use a loop for both files instead of copypasting them and maintaining both separately.

Helenerineium:
Mostly fixed, please review again.

wiiaboo:
Better, but I still disagree with calling the first commit `static linking` as this is nothing but. Reword it to be `Vapoursynth: use delay-import lib, still depends on shared libs`.;

Helenerineium:
@wiiaboo Commit message updated.

wiiaboo:
Yeah, I was waiting for the weekend to do a final test compile before approving it.

wiiaboo:
Merged in 92e11a834af69ebb5841c96120ccd24e8b2fee31

PR #1239, opened 2019-06-05
bat: export CXX

schmidthubert:
Weeks ago I had a problem with libplacebo. While running a mingw shell libplacebo compiled just fine but the script errored. Going to the bottom of it revealed it was trying to use cl.exe from orig_path in Windows (which we use here) instead of g++/c++. After trying things like `-DCMAKE_CXX_COMPILER=$(command -v g++.exe)` and or saving and restoring PATHs like `#PATH=${MINGW_PREFIX}/bin:${MSYS2_PATH}`  etc I just made it global (as it is set in some LX distros as well) and never had a problem since then. So I'm PR'ing this now.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1626,9 +1626,10 @@ goto :EOF
     echo.PKG_CONFIG_PATH="${LOCALDESTDIR}/lib/pkgconfig:${MINGW_PREFIX}/lib/pkgconfig"
     echo.CPPFLAGS="-D_FORTIFY_SOURCE=2 -D__USE_MINGW_ANSI_STDIO=1"
     echo.CFLAGS="-mthreads -mtune=generic -O2 -pipe"
+    echo.CXX="g++"
     echo.CXXFLAGS="${CFLAGS}"
     echo.LDFLAGS="-pipe -static-libgcc -static-libstdc++"
-    echo.export DXSDK_DIR ACLOCAL_PATH PKG_CONFIG PKG_CONFIG_PATH CPPFLAGS CFLAGS CXXFLAGS LDFLAGS MSYSTEM
+    echo.export DXSDK_DIR ACLOCAL_PATH PKG_CONFIG PKG_CONFIG_PATH CPPFLAGS CFLAGS CXX CXXFLAGS LDFLAGS MSYSTEM
     echo.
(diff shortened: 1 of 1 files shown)


wiiaboo:
Does it still happen now that we're only defining search paths within mingw_prefix and suite_prefix? Unless it's more retarded CMake behavior, it should not be trying to look for programs outside of those prefixes.

schmidthubert:
`echo.PATH="${LOCALDESTDIR}/bin:${MINGW_PREFIX}/bin:${INFOPATH}:${MSYS2_PATH}:${ORIGINAL_PATH}"`

is still there. And it doesn't have anything to do with cmake. It wants to do c++ but somehow uses cl.exe from Windows. If I use 

`OLD_PATH="$PATH"`
PATH=${MINGW_PREFIX}/bin:${MSYS2_PATH}
`PATH="$OLD_PATH"`

it works as well. But CXX env var should be global, why fiddle with PATH locally?

1480c1:
Cmake should not choose cl.exe for any generators other than vs or manual set, is it possible libplacebo is setting cxx manually if windows?

wiiaboo:
Right, libplacebo uses meson. So maybe the same prefix searching restrictions need to be done to it to avoid looking for stuff outside of the suite.

wiiaboo:
CMake seems to work fine with the recent changes, so only do_meson needs the hardcoded `CC` and `CXX` environment variables, not the whole suite.

schmidthubert:
That's not good enough for me I guess, but I do whatever I want anyway, hmm! But you should really checkout some distros ENV vars, VMs etc what's set there, there's nothing at all wrong with setting this. However, I'll even give you and c some extra work here: "I don't like here ..." or better: (and as you prefer it) "opening a MABS shell" doesn't produce the same results AT ALL as if the script is running/called from the batch . This should be worked on and improved. :+1: 

So I got 0 points here for my todays posting here for anything, that's very interesting.

wiiaboo:
> But you should really checkout some distros ENV vars, VMs etc what's set there, there's nothing at all wrong with setting this. 

I know other people do set these. That doesn't mean I still don't think it's a good practice to force it globally when the other libs are finding them correctly. I'd maybe rather have only FFmpeg getting the PATH polluted with cl.exe if needed.

> So I got 0 points here for my todays posting here for anything, that's very interesting.

It's not a competition. End result is things working, not who's committing more or doing more work.

PR #1249, opened 2019-06-14
svthevc: new transition patch due to changes

1480c1:
New patch due to changes in 213

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1227,7 +1227,7 @@ _check=(SvtHevcEnc.pc libSvtHevcEnc.dll.a svt-hevc/EbApi.h
 if [[ $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
     do_uninstall bin-video/SVT-Hevc/libSvtHevcEnc.dll.a "${_check[@]}"
     do_patch "https://github.com/OpenVisualCloud/SVT-HEVC/pull/213.patch"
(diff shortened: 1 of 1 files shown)


1480c1:
addresses https://github.com/jb-alvarado/media-autobuild_suite/issues/1103#issuecomment-502077968

wiiaboo:
If you foresee more changes soon/before the PRs are merged, you could also use gist instead, so the URL stays the same and you don't need to keep changing this.

1480c1:
I might as well

1480c1:
Done

wiiaboo:
Hm, actually that'll still be different. The second hash-looking path parameter seems like it'd be a commit hash.

build/media-suite_compile.sh
```diff
@@ -1227,7 +1227,7 @@ _check=(SvtHevcEnc.pc libSvtHevcEnc.dll.a svt-hevc/EbApi.h
 if [[ $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
     do_uninstall bin-video/SVT-Hevc/libSvtHevcEnc.dll.a "${_check[@]}"
     do_patch "https://github.com/OpenVisualCloud/SVT-HEVC/pull/213.patch"
-    do_patch "http://0x0.st/zQP4.patch"
+    do_patch "https://gist.githubusercontent.com/1480c1/a5fbbff6aa2cb9880275ae6ad63bd9df/raw/59922fb6a7b242627dea517acc6d21dae6b9aee2/0001-transition-between-y4m-and-cmakelists.patch"
```
wiiaboo:
```suggestion
    do_patch "https://gist.githubusercontent.com/1480c1/a5fbbff6aa2cb9880275ae6ad63bd9df/raw/0001-transition-between-y4m-and-cmakelists.patch"
```
If you remove the commit path parameter it seems to point to the HEAD, it seems.

PR #1250, opened 2019-06-16
x265: add svt-hevc

### LINKED ISSUE #1103: request: x265 with SVT-HEVC

1480c1:
Add svt-hevc to x265, Closes #1103

Also try to simplify x265 building a little bit.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1362,13 +1362,11 @@ if [[ ! $x265 = "n" ]] && do_vcs "hg::https://bitbucket.org/multicoreware/x265";
     do_patch "https://gist.githubusercontent.com/1480c1/7743056f52e4a546294e2160431674bc/raw/0001-CMake-alias-CMAKE_INSTALL_BINDIR-to-BIN_INSTALL_DIR.patch"
 
     do_x265_cmake() {
-        do_print_progress "Building $1" && shift 1
-        extra_script pre cmake
-        log "cmake" cmake "$LOCALBUILDDIR/$(get_first_subdir)/source" -G Ninja \
-        -DCMAKE_INSTALL_PREFIX="$LOCALDESTDIR" -DBIN_INSTALL_DIR="$LOCALDESTDIR/bin-video" \
+        do_print_progress "Building $1" && shift
+        do_cmake video "$(realpath --relative-to="$PWD" "$LOCALBUILDDIR/$(get_first_subdir)/source")" \
+        skip_build_dir
         -DENABLE_SHARED=OFF -DENABLE_CLI=OFF -DHIGH_BIT_DEPTH=ON -DHG_EXECUTABLE=/usr/bin/hg.bat \
         -DENABLE_HDR10_PLUS=ON $xpsupport "$@"
-        extra_script post cmake
         do_ninja
     }
     build_x265() {
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1359,6 +1359,7 @@ if [[ ! $x265 = "n" ]] && do_vcs "hg::https://bitbucket.org/multicoreware/x265";
     do_uninstall libx265{_main10,_main12}.a bin-video/libx265_main{10,12}.dll "${_check[@]}"
     [[ $bits = "32bit" ]] && assembly="-DENABLE_ASSEMBLY=OFF"
     [[ $x265 = d ]] && xpsupport="-DWINXP_SUPPORT=ON"
+    do_patch "https://gist.githubusercontent.com/1480c1/7743056f52e4a546294e2160431674bc/raw/0001-CMake-alias-CMAKE_INSTALL_BINDIR-to-BIN_INSTALL_DIR.patch"
 
     build_x265() {
         create_build_dir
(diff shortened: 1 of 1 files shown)


1480c1:
Question: what is the point of ` [[ -d "./$1" ]] && root="../$1"` in 
https://github.com/jb-alvarado/media-autobuild_suite/blob/da1c7f02a32839b1a72f05be711c0feba0949cc4/build/media-suite_helper.sh#L1090

Should it not be ` [[ -d "./$1" ]] && root="./$1"` since we are checking if the dir exists relative to current rather than parent?

1480c1:
> Question: what is the point of ` [[ -d "./$1" ]] && root="../$1"` in
> https://github.com/jb-alvarado/media-autobuild_suite/blob/da1c7f02a32839b1a72f05be711c0feba0949cc4/build/media-suite_helper.sh#L1090
> 
> Should it not be ` [[ -d "./$1" ]] && root="./$1"` since we are checking if the dir exists relative to current rather than parent?

Wait, I realized, that's relative to the created build dir, nvm

wiiaboo:
> > Should it not be ` [[ -d "./$1" ]] && root="./$1"` since we are checking if the dir exists relative to current rather than parent?
> 
> Wait, I realized, that's relative to the created build dir, nvm

Yeah, using relative paths there is kind of confusing.

build/media-suite_helper.sh
```diff
@@ -1083,24 +1083,36 @@ do_custom_patches() {
...
-    create_build_dir
+    [[ -z "$skip_build_dir" ]] && create_build_dir $cmake_build_dir
     extra_script pre cmake
     [[ -f "$(get_first_subdir)/do_not_reconfigure" ]] &&
         return
     log "cmake" cmake "$root" -G Ninja -DBUILD_SHARED_LIBS=off \
         -DCMAKE_TOOLCHAIN_FILE="$LOCALDESTDIR/etc/toolchain.cmake" \
         -DCMAKE_INSTALL_PREFIX="$LOCALDESTDIR" -DUNIX=on \
-        -DCMAKE_BUILD_TYPE=Release $bindir "$@"
+        -DCMAKE_BUILD_TYPE=Release $bindir ${cmake_extras[@]} "$@"
```
wiiaboo: Where's this variable coming from?

1480c1: That's for the user to use in extra_*.sh if they want

1480c1:
Unsure if there was anything else I needed to add, only need to test now

build/media-suite_compile.sh
```diff
@@ -1359,23 +1359,27 @@ if [[ ! $x265 = "n" ]] && do_vcs "hg::https://bitbucket.org/multicoreware/x265";
     do_uninstall libx265{_main10,_main12}.a bin-video/libx265_main{10,12}.dll "${_check[@]}"
     [[ $bits = "32bit" ]] && assembly="-DENABLE_ASSEMBLY=OFF"
     [[ $x265 = d ]] && xpsupport="-DWINXP_SUPPORT=ON"
+    do_patch "https://gist.githubusercontent.com/1480c1/7743056f52e4a546294e2160431674bc/raw/0001-CMake-alias-CMAKE_INSTALL_BINDIR-to-BIN_INSTALL_DIR.patch"
```
wiiaboo: Did you submit it upstream?

wiiaboo:
You could also remove your merged patch in svthevc in this same PR.

build/media-suite_compile.sh
1480c1: Not entirely sure how to do pr and stuff with hg and bitbucket. I want to, but still need to learn how to use those

wiiaboo: Sure, as long as you're thinking of doing it. Unless a project is dead or the patch would be considered unmergeable (like FFmpeg's configure patch), it's better to consider them temporary.

1480c1: Right now reading their CLA and trying to figure out how to use patchwork

wiiaboo: Don't forget to check whether they're using GNUInstallDirs already. If not, then the patch(es) would need to be more than just using CMAKE_INSTALL_BINDIR, or it could be refused.

1480c1: I think they weren't using GNUInstallDirs, that's why I added in the patch
``` patch
+if(NOT CMAKE_INSTALL_BINDIR)
+    set(CMAKE_INSTALL_BINDIR bin)
+endif()
```

1480c1: My main problem is that since they have a minimum of 2.8.8 for cmake, I need to make sure whatever I write supports 2.8.8, but the cmake docs online only go to 3.0.2

wiiaboo: You could probably propose updating the minimum version too. How old is 3.0.2 already? There's absolutely no reason why versions older than one or two years should be supported (or the latest version in an Ubuntu LTS usually).

wiiaboo: Even old-ass xenial (Ubuntu 16.04 LTS) has 3.5.1. Propose an update to that version, at the very least.

1480c1: I will try, 3.5 is at least a pretty good version to start with

1480c1: It seems centos 7 still uses 2.8.12, holy hell their packages are out of date

wiiaboo: If you're building stuff from hg/git you can also not use a CMake from 10 years ago. Aim for 3.5.1, at least.

1480c1: Probably best to forget that centos 7 exists

PR #1251, opened 2019-06-16
svthevc: don't use march=native flag

1480c1:
Forgot I had added a NATIVE flag for using march=native


wiiaboo:
tbh, this shouldn't even be an option. If people want to use `-march=native`, they'll set it on their CFLAGS/CXXFLAGS.

PR #1252, opened 2019-06-16
do_cmake: add more variables and switches

1480c1:
Allows for extra_script to add flags for cmake
Allows for option to skip `create_build_dir` or pass a name to `create_build_dir`


build/media-suite_helper.sh
```diff
@@ -1083,24 +1083,36 @@ do_custom_patches() {
...
+            else
+                bindir="$1"
+            fi
+            shift && break ;;
+        esac
+    done
 
     local PKG_CONFIG="$LOCALDESTDIR/bin/ab-pkg-config-static.bat"
-    create_build_dir
+    [[ -z "$skip_build_dir" ]] && create_build_dir $cmake_build_dir
```
wiiaboo: Shouldn't `create_build_dir` always run, though? Or is this specifically for in-root builds?

1480c1: This is for the builds where a build for is already created, just not with `build*_-64-bits`

wiiaboo: Oh, right, sure.

PR #1258, opened 2019-06-18
svthevc: add SVT-HEVC to ffmpeg

1480c1:
Needs to be tested as the patches are based off the 4.1 branch

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1709,6 +1709,13 @@ if [[ $ffmpeg != "no" ]]; then
             do_patch "https://0x0.st/zeB6.txt"
         fi
 
+        if enabled libsvthevc; then
+            SVT_HEVC_ffmpeg_patches="https://raw.githubusercontent.com/OpenVisualCloud/SVT-HEVC/master/ffmpeg_plugin"
+            do_patch "${SVT_HEVC_ffmpeg_patches}/0001-lavc-svt_hevc-add-libsvt-hevc-encoder-wrapper.patch" am
+            do_patch "${SVT_HEVC_ffmpeg_patches}/0002-doc-Add-libsvt_hevc-encoder-docs.patch" am
+            unset SVT_HEVC_ffmpeg_patches
+        fi
+
         # shared
         if [[ $ffmpeg != "static" ]] && [[ ! -f build_successful${bits}_shared ]]; then
             do_print_progress "Compiling ${bold}shared${reset} FFmpeg"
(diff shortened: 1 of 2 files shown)


1480c1:
Although I do see the patches at patchwork from jsunintel, I'm not sure if it would be a good idea to use those mainly because those seem to be updated a lot

wiiaboo:
I think it's better to check if ffmpeg_options.txt contains `--enable-libsvthevc` instead than always adding it with `svthevc=yes` (and apply the patch only if the option is there, or remove it if the patch fails to apply to prevent configure breakage). People might want the standalone encoder on its own or included in x265 but not include it in FFmpeg.

1480c1:
Okay, should I leave a note on the bat file to add the flag if they want to use svt-hevc with ffmpeg?

wiiaboo:
Sure.

wiiaboo:
There's also the ffmpeg_options_64.txt hack, but that's replacing the common one, not incrementing it.

build/media-suite_compile.sh
```diff
@@ -1224,7 +1224,7 @@ fi
 
 _check=(SvtHevcEnc.pc libSvtHevcEnc.dll.a svt-hevc/EbApi.h
     bin-video/{libSvtHevcEnc.dll,SvtHevcEncApp.exe})
-if [[ $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
+if [[ $bits != "32bits" && $svthevc = "y" ]] && do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
```
wiiaboo: It's `32bit` actually. But might as well use `$bits = "64bit"`.

wiiaboo:
```suggestion
if [[ $bits = "32bit" ]]; then
    do_removeOption --enable-libsvthevc
elif { [[ $svthevc = "y" ]] || enabled libsvthevc; } &&
    do_vcs "https://github.com/OpenVisualCloud/SVT-HEVC.git"; then
```

build/media-suite_compile.sh
```diff
@@ -1709,6 +1709,13 @@ if [[ $ffmpeg != "no" ]]; then
             do_patch "https://0x0.st/zeB6.txt"
         fi
 
+        if enabled libsvthevc && [[ $bits != "32bits"]]; then
```
wiiaboo: Then you can leave just the check for `enabled libsvthevc` here.

1480c1: Done

PR #1267, opened 2019-06-26
Activate libsvthevc for ffmpeg_options_full

BlackScreen:
Since the building process is running smoothly with libsvthevc activated, why not activate it as default...

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ For information about the compiler environment see the wiki, there you also have
         - librubberband (git snapshot)
         - libsrt (git)
         - libssh (mingw)
+        - libsvthevc (git)
         - libtesseract (git)
         - libvmaf (git)
         - libxavs (svn snapshot)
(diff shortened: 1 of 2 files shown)


1480c1:
Because the user might not have svt-hevc enabled

1480c1:
If you are going to do that, try to see if you can make sure there is logic to remove the flag if svthevc is not set to y

wiiaboo:
That's why I asked to add this second case in svthevc's if: https://github.com/jb-alvarado/media-autobuild_suite/blob/master/build/media-suite_compile.sh#L1229

Unlike for x264, x265, and some others, svthevc will still be built if only FFmpeg's option is enabled.

BlackScreen:
Ok, if it is not a good idea, then close this case. :)

1480c1:
It seems fine since It seems that if libsvthevc is enabled, it will build svthevc

wiiaboo:
Merged in 01fe9e7e3bbfe22783a6e576167752af4156bba7

PR #1273, opened 2019-07-05
OpenCL-Headers: use latest version from git

Helenerineium:


wiiaboo:
Does this make any difference? Won't this lead to more broken builds with older drivers?

Helenerineium:
I don't know. The opencl-headers package provided by msys2 are version 2.2 specific headers.

But the headers in git can target older version by defining `CL_TARGET_OPENCL_VERSION` (default 220 - OpenCL 2.2)

wiiaboo:
OK, if someone's build gets broken we can test that later.

PR #1278, opened 2019-07-08
Use download names for patches from Patchwork

lars18th:
When donwloading a patch is mandatory to use a filename for the download patch. Without it the patch fails. So, show a correct example.


wiiaboo:
Why?

PR #1291, opened 2019-07-25
Helper: Remove unecessary ncolors variable

1480c1:
Test tput directly.

Is there anything else that requires the ncolors variable?


wiiaboo:
Not that I know of.

PR #1292, opened 2019-07-25
do_print: Use print's variable setting and simplify printing of status

1480c1:
Maybe we should have a style guideline.


wiiaboo:
Merged in 5e4e789

PR #1293, opened 2019-07-25
Escapes: Use single quotes for some, combine two echos into printf

1480c1:
Does echo even have an exit code other than zero?


1480c1:
For 629ce8899aa0e29aa96a7348038cc27f57639ab5, which do you think would be better? `find | xargs rm` or `find exec rm and find delete` ?

wiiaboo:
If delete and exec rm are functionally the same as xargs rm, the former are fine by me.

1480c1:
`-delete` will act like rm without r, so it will work with single files, but not directories, `-exec rm -rf {} +` will work like xargs without the additional call to xargs itself

PR #1294, opened 2019-07-25
Helper: Redo some logic for check_hash and real_extract and related

1480c1:
move the logic for extracting into real_extract. (Now works on linux.)

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -295,11 +295,11 @@ guess_dirname() {
 check_hash() {
     local file="$1" check="$2" md5sum sha256sum
     if [[ -f $file ]]; then
-        md5sum=$(md5sum "$file" | awk '{ print $1 }')
-        sha256sum=$(sha256sum "$file" | awk '{ print $1 }')
+        sha256sum=$(sha256sum "$file" | cut -d' ' -f1)
         if [[ $check = print ]]; then
             echo "$sha256sum"
         else
+            md5sum=$(md5sum "$file" | cut -d' ' -f1)
             test "$sha256sum" = "$check" || test "$md5sum" = "$check"
         fi
     else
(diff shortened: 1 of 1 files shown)


1480c1:
Seems to work now

build/media-suite_helper.sh
```diff
@@ -372,20 +372,30 @@ do_wget() {
...
+            7z x -aoa "${archive%.tar*}.tar" -o"$dirName"
         fi
-        rm -f "${1%.tar*}.tar"
+        rm -f "${archive%.tar*}.tar"
         ;;
     esac
+    local temp_dir
+    temp_dir=$(find "$dirName/" -maxdepth 1 ! -wholename "$dirName/")
+    if [[ -n "$temp_dir" && $(wc -l <<< "$temp_dir" ) = 1 ]]; then
+        find "$temp_dir" -maxdepth 2 ! -wholename "$temp_dir" -exec mv -t "$dirName/" {} +
```
wiiaboo: Should be `-maxdepth 1` otherwise you'll be trying to move files inside root-level directories after moving the directories themselves.

1480c1: I have a feeling that's why I added `! -wholename` but I'll check again

1480c1: You were right about the max depth

wiiaboo: No, `! -wholename "$temp_dir/"` just ignores/skips the own "$temp_dir" entry.

1480c1:
I'll go through each call of do_wget and others that use real_extract and confirm that they all work

build/media-suite_helper.sh
wiiaboo: Which makes sense, but you don't need to explicitely move files from inside root-level directories, as they're already moved when moving the directories.

wiiaboo:
Probably just need to check zip/7z/rar as `.tar.*` are usually missing a root-level directory.

1480c1:
Yea, main reason why I even added the checking of redundant dirs is to make sure that the dirname is always used

1480c1:
Going to apply once I'm properly on my computer, has to do it properly through a phone screen

wiiaboo:
I've rebased+merged in 5bb087198ddef3513ad560ce6103c1bf2ef890d3.

PR #1302, opened 2019-07-26
NoMintty: Change how compile.sh fails

1480c1:
Using Start-Process caused the cursor to reset to top left

Use declare -p since that doesn't output functions


wiiaboo:
Sure? Who was it that requested this noMintty, so they can test it?

For the rest of your PRs, I'll try to go over them this weekend.

1480c1:
Lars, but I also use it to so I can still see the output of the script without it disappearing with mintty and without looking in the compile log

PR #1303, opened 2019-07-29
do_simple_print: Add wrapper for timeStamp for printing messages

1480c1:
Going to test now.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -71,11 +71,11 @@ done
 source "$LOCALBUILDDIR"/media-suite_helper.sh
 
 [[ -f "$LOCALBUILDDIR/no_logs" ||  $logging = n ]] &&
-    echo -e "${orange}Warning: We will not accept any issues lacking any form of logs or logs.zip!${reset}"
+    do_simple_print -p "${orange}Warning: We will not accept any issues lacking any form of logs or logs.zip!${reset}"
 
 buildProcess() {
 set_title
-echo -e '\n\t'"${orange}Starting $bits compilation of all tools${reset}"
+do_simple_print -p '\n\t'"${orange}Starting $bits compilation of all tools${reset}"
 [[ -f "$HOME"/custom_build_options ]] &&
     echo "Imported custom build options (unsupported)" &&
     source "$HOME"/custom_build_options
@@ -133,7 +133,7 @@ create_ab_pkgconfig
 create_cmake_toolchain
 
 set_title "compiling global tools"
-echo -e '\n\t'"${orange}Starting $bits compilation of global tools${reset}"
+do_simple_print -p '\n\t'"${orange}Starting $bits compilation of global tools${reset}"
 
 if [[ $packing = y ]] &&
     ! [[ -e /opt/bin/upx.exe && "$(/opt/bin/upx -V | head -1)" = "upx 3.95" ]] &&
@@ -526,7 +526,7 @@ if [[ $ffmpeg != "no" ]] && enabled libzimg &&
 fi
 
 set_title "compiling audio tools"
-echo -e '\n\t'"${orange}Starting $bits compilation of audio tools${reset}"
+do_simple_print -p '\n\t'"${orange}Starting $bits compilation of audio tools${reset}"
 
 if [[ $ffmpeg != "no" || $sox = y ]]; then
     enabled libwavpack && do_pacman_install wavpack
@@ -866,7 +866,7 @@ if { { [[ $ffmpeg != "no" ]] &&
 fi
 
 set_title "compiling video tools"
-echo -e '\n\t'"${orange}Starting $bits compilation of video tools${reset}"
+do_simple_print -p '\n\t'"${orange}Starting $bits compilation of video tools${reset}"
 
 _deps=(gnutls.pc)
 _check=(librtmp.{a,pc})


build/media-suite_compile.sh
```diff
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2034,SC1090,SC1117,SC1091,SC2119
 shopt -s extglob
 
-FFMPEG_BASE_OPTS=(--pkg-config-flags=--static)
+FFMPEG_BASE_OPTS=("--pkg-config-flags=--static")
```
wiiaboo: What's the issue with these?

build/media-suite_compile.sh
```diff
@@ -1572,7 +1575,7 @@ if  { ! mpv_disabled vapoursynth || enabled vapoursynth; }; then
             gendef - "../$_file.dll" >/dev/null 2>&1 |
                 sed -r -e 's|^_||' -e 's|@[1-9]+$||' > "${_file}.def"
             dlltool -y "lib${_file}.a" -d "${_file}.def" \
-                $([[ $bits = 32bit ]] && echo "-U") 2>/dev/null
+                "$([[ $bits = 32bit ]] && echo "-U")" 2>/dev/null
```
wiiaboo: Isn't this going to be an empty parameter if `$bits = 64bit`?

build/media-suite_compile.sh
1480c1: Shellcheck keeps complaining about the = being literal due to thinking we are trying to use associative arrays

build/media-suite_compile.sh
1480c1: I think so too, might need a better way or just ignore the "prevent word splitting" warning

wiiaboo: Check these case-by-case. Most cases don't like empty parameters, like in this case. This needs another way of improving it, or just ignore the issue.

build/media-suite_compile.sh
```diff
@@ -1570,9 +1570,9 @@ if  { ! mpv_disabled vapoursynth || enabled vapoursynth; }; then
             [vsscript-description]='Library for interfacing VapourSynth with Python'
             [vsscript-private]="-l$_python_lib -lstdc++"
         )
-        for _file in vapoursynth vsscript; do
-            gendef - "../$_file.dll" >/dev/null 2>&1 |
-                sed -r -e 's|^_||' -e 's|@[1-9]+$||' > "${_file}.def"
+        for _file in VapourSynth VSScript; do
```
wiiaboo: Be careful with these. Most unix tools expect case sensitivity, Windows doesn't mind it.

build/media-suite_compile.sh
1480c1: Yea, going to just ignore it since dlltool doesn't seem to like it

build/media-suite_compile.sh
1480c1: Checked the original archive, the those are the actual cases.

1480c1: I was confused when I tried to tab autocomplete the name, but nothing came up

wiiaboo: I mean in the sense that FFmpeg's configure and mpv's configure will use the all lowercase names to look for the libnames. `gendef` and `dlltool` don't mind being given wrong-case files, since they're Windows tools.
Either that or, if you want to make code that's "linux-equivalent", use the all-lowercase to generate the new files, since that's the canonical name (same reason why vsscript-name is vapoursynth-script).

1480c1: I see, I see, going to just revert this part since adding another variable just for lowercased file name is probably not worth it at this point.

Maybe in the future if I decide to endeavor to make the suite work on linux. That's probably a long time away

1480c1:
I may need to relook into the using git mirror since it does not seem to work properly

1480c1:
Accidentally mislabeled some commits with helper instead of update

build/media-suite_compile.sh
```diff
@@ -1760,12 +1760,12 @@ if [[ $ffmpeg != "no" ]]; then
 
         if [[ $ffmpeg = "both" ]]; then
             _check+=(bin-video/ffmpegSHARED/lib/"libavutil${build_suffix}.dll.a")
-            FFMPEG_OPTS_SHARED+=(--prefix="$LOCALDESTDIR/bin-video/ffmpegSHARED")
+            FFMPEG_OPTS_SHARED+=("--prefix=\"$LOCALDESTDIR/bin-video/ffmpegSHARED\"")
```
wiiaboo: Doesn't `"--prefix=$LOCALDESTDIR/bin-video/ffmpegSHARED"` work just as well, like above?

build/media-suite_compile.sh
wiiaboo: Still haven't removed the quotes here.

build/media-suite_compile.sh
1480c1: I would think so too, but at the same time, I'm slightly concerned if we should account for people who might try to redirect $LOCALDESTDIR to a separate folder, or should we just not care about those kinds of cases?

build/media-suite_compile.sh
1480c1: I don't think the inside echo quotes would matter

build/media-suite_compile.sh
wiiaboo: Fuck them.

build/media-suite_compile.sh
wiiaboo: Nevermind, I didn't see it was fixed in a later commit.

build/media-suite_compile.sh
wiiaboo: Nevermind, I didn't see it was fixed in a later commit.

build/media-suite_compile.sh
wiiaboo: Yeah, this should be removed. Literal quotes will break FFmpeg's configure.

1480c1: Removed

wiiaboo:
Cherry-picked most commits, just missing the simple_print, as I haven't tested that yet.

1480c1:
do_simple_print is just meant to be a wrapper for timeStamp and making it so that those individual messages are so jarringly placed on the left side

wiiaboo:
Can you just add "$reset" directly to simple_print? I don't think there's any reason why it shouldn't always be there at the end of the line.

1480c1:
Sure

1480c1:
after `$*` or inside the printf format?

wiiaboo:
Maybe cleaner inside the printf format string?

build/media-suite_helper.sh
```diff
@@ -22,6 +22,24 @@ ncols=72
...
     rm -f "$LOCALBUILDDIR"/{7za,wget,grep}.exe
 
+do_simple_print() {
+    local plain
+    [[ $1 = -p ]] && plain=y && shift
+    if [[ $timeStamp = y ]]; then
+        if [[ $plain = y ]]; then
+            printf "${purple}"'%(%H:%M:%S)T'"${reset}"' %b\n' -1 "$*"
+        else
+            printf "${purple}"'%(%H:%M:%S)T'"${reset} ${bold}├${reset}"' %b\n' -1 "$*"
```
wiiaboo: What's the `-1` for?

1480c1: the time

wiiaboo: Ah, nvm, it's needed for the date.

wiiaboo:
```bash
do_simple_print() {
    local plain formatString='' dateValue
    [[ $1 = -p ]] && plain=y && shift

    if [[ $timeStamp = y ]]; then
        formatString="${purple}"'%(%H:%M:%S)T'"${reset}"' '
        dateValue='-1'
    else
        [[ $plain = y ]] && formatString='\t'
    fi
    if [[ -z $plain ]]; then
        formatString+="${bold}├${reset} "
    fi
    printf "$formatString"'%b\n' $dateValue "$*"
}
```
Does this look simpler or more confusing?

1480c1:
That actually looks pretty good

wiiaboo:
Merged as ed9b52e. Fixedup the missing hardcoded `$reset`.

wiiaboo:
Hm, there's a few `echo -e` left. Sorry.

1480c1:
I left the `Starting $bits compilation...` alone since I'm not sure if they would look good with do_simple_print, Should I change those too?

1480c1:
going to test it out and see how they look

1480c1:
``` none
License: nonfree
Imported options from mpv_options.txt
17:41:57
        Starting 64bit compilation of global tools
17:41:57 ├ Running rustup_update...
17:41:57 ├ Running set_default_toolchain...
17:41:59   Running git update for ripgrep...
17:41:59 ripgrep git  ...................................... [Up-to-date]
```

I guess it looks ok

PR #1305, opened 2019-07-30
do_pacman: Rework do_pacman_* to accept msys2 packages

1480c1:
Rework to accept msys2 packages and remove extra calls to grep

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1438,12 +1438,17 @@ do_unhide_all_sharedlibs() {
 }
 
 do_pacman_install() {
-    local installed
-    local pkg
-    installed="$(pacman -Qqe | /usr/bin/grep "^${MINGW_PACKAGE_PREFIX}-")"
+    local pkg msyspackage
+    while [ -n "$*" ]; do
+        case "$1" in
+        -m) msyspackage=y ;;
+        *) break ;;
+        esac
+    done
     for pkg; do
-        [[ "$pkg" != "${MINGW_PACKAGE_PREFIX}-"* ]] && pkg="${MINGW_PACKAGE_PREFIX}-${pkg}"
-        /usr/bin/grep -q "^${pkg}$" <(echo "$installed") && continue
+        [[ $msyspackage != "y" && "$pkg" != "${MINGW_PACKAGE_PREFIX}-"* ]] &&
+            pkg="${MINGW_PACKAGE_PREFIX}-${pkg}"
+        pacman -Qqe "^${pkg}$" >/dev/null 2>&1 && continue
         if [[ $timeStamp = y ]]; then
             printf "${purple}"'%(%H:%M:%S)T'"${reset}"' %s' -1 "Installing ${pkg#$MINGW_PACKAGE_PREFIX-}... "
         else
@@ -1453,6 +1458,7 @@ do_pacman_install() {
             pacman -D --asexplicit "$pkg" >/dev/null
             /usr/bin/grep -q "^${pkg#$MINGW_PACKAGE_PREFIX-}$" /etc/pac-mingw-extra.pk >/dev/null 2>&1 ||
                 echo "${pkg#$MINGW_PACKAGE_PREFIX-}" >> /etc/pac-mingw-extra.pk
+            sort -uo /etc/pac-mingw-extra.pk{,}
             echo "done"
         else
             echo "failed"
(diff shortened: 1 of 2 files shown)


wiiaboo:
How does this now accept msys2 packages? Looks to be doing the same thing.
Also, this one is pretty important that there's no regressions in behavior. I'd even rather you do a new version of the function meanwhile to be absolutely sure we don't start installing the wrong packages.

1480c1:
It won't override pkg with a mingw-w64 prefix if mingw-w64 version doesn't exist

1480c1:
First regression would probably be in the update script

wiiaboo:
There's a lot of possible places for regressions with this one.

1480c1:
That is true

build/media-suite_helper.sh
```diff
@@ -1419,7 +1419,8 @@ do_pacman_install() {
             pacman -D --asexplicit "$pkg" >/dev/null
             /usr/bin/grep -q "^${pkg#$MINGW_PACKAGE_PREFIX-}$" /etc/pac-mingw-extra.pk >/dev/null 2>&1 ||
                 echo "${pkg#$MINGW_PACKAGE_PREFIX-}" >> /etc/pac-mingw-extra.pk
-            echo "done"
+            sort -uo /etc/pac-mingw-extra.pk{,}
+            echo "done" 2>/dev/null
```
wiiaboo: What's the redirection for?

wiiaboo:
I think I still would prefer that this doesn't ever install msys2 packages. It should be an explicit option at the very least, and copy the package names to a different file, not `/etc/pac-mingw-extra.pk`.
The rest of the optimizations are fine, though I'm not a big fan of the extra pacman calls.

build/media-suite_helper.sh
1480c1: Not entirely sure, going to revert

1480c1:
Sure, might make it use a switch like `-m` to mark a msys2 package explicitly then

1480c1:
using `do_pacman_install msys mercurial` vs `do_pacman_install -m mercurial`, which would be better?

wiiaboo:
The `-m` one.

build/media-suite_helper.sh
```diff
@@ -1424,12 +1424,17 @@ do_unhide_all_sharedlibs() {
...
+        -m) msyspackage=y ;;
+        *) break ;;
+        esac
+    done
     for pkg; do
-        [[ "$pkg" != "${MINGW_PACKAGE_PREFIX}-"* ]] && pkg="${MINGW_PACKAGE_PREFIX}-${pkg}"
-        /usr/bin/grep -q "^${pkg}$" <(echo "$installed") && continue
+        [[ $msyspackage != "y" && "$pkg" != "${MINGW_PACKAGE_PREFIX}-"* ]] &&
+            pkg="${MINGW_PACKAGE_PREFIX}-${pkg}"
+        pacman -Qqe "$pkg" >/dev/null 2>&1 && continue
```
wiiaboo: If you remove the ^$ anchors this will match with packages that include the name you're looking for.
i.e. do_pacman_install python will match `python3` and `mingw-w64-x86_64-python`, `mingw-w64-x86_64-python3-docutils`, etc.

build/media-suite_helper.sh
```diff
@@ -1448,14 +1454,20 @@ do_pacman_install() {
...
+    done
     for pkg; do
-        [[ "$pkg" != "${MINGW_PACKAGE_PREFIX}-"* ]] && pkg="${MINGW_PACKAGE_PREFIX}-${pkg}"
+        [[ $msyspackage != "y" && "$pkg" != "${MINGW_PACKAGE_PREFIX}-"* ]] &&
+            pkg="${MINGW_PACKAGE_PREFIX}-${pkg}"
         [[ -f /etc/pac-mingw-extra.pk ]] &&
             sed -i "/^${pkg#$MINGW_PACKAGE_PREFIX-}$/d" /etc/pac-mingw-extra.pk >/dev/null 2>&1
-        /usr/bin/grep -q "^${pkg}$" <(echo "$installed") || continue
+        sort -uo /etc/pac-mingw-extra.pk{,}
+        pacman -Qqe "$pkg" >/dev/null 2>&1 || continue
```
wiiaboo: Same.

wiiaboo:
If it's installing an msys package it should maybe save the name to a different file as well, not pac-mingw-extra.pk but pac-msys-extra.pk, but I've said this before.

1480c1:
Done

build/media-suite_helper.sh
```diff
@@ -1434,17 +1434,23 @@ do_pacman_install() {
...
         if [[ $timeStamp = y ]]; then
             printf "${purple}"'%(%H:%M:%S)T'"${reset}"' %s' -1 "Installing ${pkg#$MINGW_PACKAGE_PREFIX-}... "
         else
             echo -n "Installing ${pkg#$MINGW_PACKAGE_PREFIX-}... "
         fi
         if pacman -S --force --noconfirm --needed "$pkg" >/dev/null 2>&1; then
             pacman -D --asexplicit "$pkg" >/dev/null
-            /usr/bin/grep -q "^${pkg#$MINGW_PACKAGE_PREFIX-}$" /etc/pac-mingw-extra.pk >/dev/null 2>&1 ||
-                echo "${pkg#$MINGW_PACKAGE_PREFIX-}" >> /etc/pac-mingw-extra.pk
+            if [[ $msyspackage != "y" ]]; then
```
wiiaboo: Isn't it the opposite?

wiiaboo:
Can you test this with the docker stuff? I won't have a Windows pc all this week.

1480c1:
Sure, I can try again with docker. I just need to disable x265 for now until I can figure out why mercurial freezes on docker

1480c1:
Using set -x and testing by removing and adding new packages, it seems that the update script works fine by asking to install and remove with mingw packages and adding packages to msys-extra works also

ImportTaste:
I didn't try this with a fresh msys2, but it is working with my existing one.

PR #1307, opened 2019-07-31
Update: Shellcheck plus rework check packages

1480c1:
To be considered before #1305, split off from #1303

Would the confirmation for missing packages really be necessary? Logically speaking, since we have marked those packages as necessary for building certain tools and codecs, wouldn't they be considered a necessary/installed later on either way?

I have also included to use shfmt this time, (caused the case spread and indention, removal of quotes from the tests, and double equal for the tests) I can remove it if it seems unnecessary

### DIFF
--- a/build/media-suite_update.sh
+++ b/build/media-suite_update.sh
@@ -48,7 +48,7 @@ if [[ "$update" = "yes" ]]; then
     echo "-------------------------------------------------------------------------------"
     echo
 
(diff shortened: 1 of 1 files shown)


1480c1:
seems to work on my system, going to try some other configurations to make sure

1480c1:
Seems to work when deleting some packages and adding in others and also when completely nuking msys64 so far

PR #1308, opened 2019-07-31
libopen264: build statically from source

Helenerineium:
Instead of updating to `openh264-2.0.0-win{32,64}.dll.bz2`, I choose to build statically from source.

This removes the dll dependency when using openh264 in ffmpeg / mpv.


wiiaboo:
No, we explicitely use the .dll, because the license doesn't work if you build your own library.

Issue #1314, opened 2019-08-03
Error when running git update for mpv


ycwan:
11:51:37   Running git update for mpv...
11:51:45 ┌ mpv git  ..................................... [Updates found]
11:51:49 ├ Running bootstrap...
11:52:30 ├ Running configure...
Likely error:
Checking for mpv CLI player                                       : yes
Checking for shared library                                       : yes
Checking for static library                                       : disabled
Checking for static build                                         : yes
Checking for whether to include binary compile time               : yes
Checking for whether to optimize                                  : yes
Checking for whether to compile-in debugging information          : disabled
Checking for manpage generation                                   : no (missing RST2MAN)
Checking for html manual generation                               : no (missing RST2HTML)
You manually enabled the feature 'html-build', but the autodetection check failed.
configure failed. Check C:/mingw/build/mpv-git/ab-suite.configure.log
This is required for other packages, so this script will exit.
11:52:48   Creating diagnostics file...

All relevant logs have been anonymously uploaded to https://0x0.st/zOr4.zip
Copy and paste [logs.zip](https://0x0.st/zOr4.zip) in the GitHub issue.
Make sure the suite is up-to-date before reporting an issue. It might've been fixed already.
Try running the build again at a later time.

Issue #1317, opened 2019-08-04
[mingw64] [mpv] missing RST2MAN


fuchanghao:
error log: https://0x0.st/zO8b.zip

Finally, the openal works now.

But mpv build fails when Check for manpage generation.

> [313m06:24:14(B[m   Running git update for mpv...
> [313m06:24:31(B[m [1m┌ mpv git (B[m .................................. [[311mRecently updated(B[m]
> [313m06:24:31(B[m [1m├(B[m Running bootstrap...
> [313m06:24:36(B[m [1m├(B[m Running configure...
> Likely error:
> Checking for internal audio filter chain                          : yes 
> Checking for mpv CLI player                                       : yes 
> Checking for shared library                                       : yes 
> Checking for static library                                       : disabled 
> Checking for static build                                         : yes 
> Checking for whether to include binary compile time               : yes 
> Checking for whether to optimize                                  : yes 
> Checking for whether to compile-in debugging information          : disabled 
> Checking for manpage generation                                   : no (missing RST2MAN) 
> You manually enabled the feature 'manpage-build', but the autodetection check failed.
> [31mconfigure failed. Check C:/media-autobuild_suite/build/mpv-git/ab-suite.configure.log(B[m
> [31mThis is required for other packages, so this script will exit.(B[m
> [313m06:24:43(B[m   Creating diagnostics file...
> 
> [32mAll relevant logs have been anonymously uploaded to https://0x0.st/zO8b.zip
> [32mCopy and paste [31m[logs.zip](https://0x0.st/zO8b.zip)[32m in the GitHub issue.(B[m
> Make sure the suite is up-to-date before reporting an issue. It might've been fixed already.
> Try running the build again at a later time.

mpv opitions.

> # Lines starting with this character are ignored
> 
> # Built-in options, use --disable- to disable them
> #--enable-cplayer
> --enable-manpage-build
> --enable-lua
> --enable-javascript
> --enable-libass
> --enable-libbluray
> --enable-uchardet
> --enable-rubberband
> --enable-lcms2
> --enable-libarchive
> --enable-libavdevice
> --enable-shaderc
> --enable-spirv-cross
> --enable-d3d11
> --enable-jpeg
> 
> # Common options or overriden defaults
> --disable-debug-build
> --lua=luajit
> 
> # Full
> --enable-dvdread
> --enable-dvdnav
> --enable-cdda
> --enable-egl-angle
> --enable-vapoursynth
> --enable-html-build
[...]

PR #1318, opened 2019-08-04
Fix rst2man/rst2html not being found

Helenerineium:
Python3 is now the default python in msys2 and don't have the '3' suffix.


Helenerineium:
Might close #1314 and #1317

PR #1342, opened 2019-08-12
do_getFFmpegConfig: strip no length options

1480c1:
experienced the same thing as https://github.com/jb-alvarado/media-autobuild_suite/issues/1339
found out that there was an extra `""` appended to the `FFMPEG_DEFAULT_OPTS`

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -937,20 +937,24 @@ disabled_all() {
 }
 
 do_getMpvConfig() {
+    local MPV_TEMP_OPTS=()
     MPV_OPTS=()
     if [[ -f "/trunk/media-autobuild_suite.bat" && "$ffmpegChoice" =~ (n|z|f) ]]; then
         IFS=$'\r\n' read -d '' -r -a bat < /trunk/media-autobuild_suite.bat
-        mapfile -t MPV_OPTS < <(do_readbatoptions "mpv_options_(builtin|basic)")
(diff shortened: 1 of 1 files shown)


wiiaboo:
Hm. Should happen with FFMPEG_OPTS_SHARED as well then?

1480c1:
Might as well since it should logically not have any detriment except for stripping ones that don't have any length.

Should I apply to mpv too? I noticed it occuring with MPV options as well

wiiaboo:
Sure. I'd never seen it happen before.

wiiaboo:
Check if these changes don't break options like --extra-ldflags="-L/mingw64 -lstuff" if they weren't broken already.

1480c1:
Testing right now and it seems that ffmpeg is fine. Only MPV failed for me earlier with the same as https://github.com/jb-alvarado/media-autobuild_suite/issues/1310

moisesmcardona:
Yup, can confirm the "" is happening again.

wiiaboo:
@moisesmcardona @LigH-de can you confirm if this PR fixes the issue?

moisesmcardona:
I'm testing it.

moisesmcardona:
Yes, it's working. Went from running configure to running make.

1480c1:
mpv works on my end

moisesmcardona:
Both ffmpeg and mpv compiled successfully.

PR #1344, opened 2019-08-13
bat: Move getting scripts for build folder up

### LINKED ISSUE #1343: wget error on new install
Error

The errors happen right after one of the initial pacman updates and the script ends. Running it again just gives the same errors.

If I run mingw64 and run pacman manually, it installs an additional 78 packages including several crypto libraries. After that update finishes, I can run the script and it will build as normal.

Annotation 2019-08-13 085653
Annotation 2019-08-13 085711

[build.zip](https://github.com/jb-alvarado/media-autobuild_suite/files/3497000/build.zip)

1480c1:
Closes #1343 

Get the scripts before running any msys2 stuff and also use powershell instead of using wget from msys2, moving reliance off of msys2 packages and onto windows

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1190,7 +1190,13 @@ if %deleteINI%==1 echo.noMintty=^%noMinttyF%>>%ini%
 ::------------------------------------------------------------------
 ::download and install basic msys2 system:
 ::------------------------------------------------------------------
-cd build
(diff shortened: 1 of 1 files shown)


wiiaboo:
Does DownloadFile support redirection?

1480c1:
I think so, it worked for me at least

PR #1354, opened 2019-08-20
Helper: Run shfmt

1480c1:
going to test to make none of this is breaking

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1838,7 +1838,7 @@ create_winpty_exe() {
 create_ab_pkgconfig() {
     # from https://stackoverflow.com/a/8088167
     local script_file
-    IFS=$'\n' read -r -d '' script_file <<'EOF' || true
+    IFS=$'\n' read -r -d '' script_file << 'EOF' || true
 #!/bin/sh
 
 while true; do
@@ -2142,7 +2142,7 @@ create_extra_skeleton() {
     [[ -f "$LOCALBUILDDIR/$extraName"_extra.sh && -z $overwrite ]] &&
         echo "$LOCALBUILDDIR/$extraName_extra.sh already exists. Use -f if you are sure you want to overwrite it." && return 1
 
-    IFS=$'\n' read -r -d '' script_file <<'EOF' || true
+    IFS=$'\n' read -r -d '' script_file << 'EOF' || true
 #!/bin/bash
 
 # Force to the suite to think the package has updates to recompile.
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1138,7 +1138,7 @@ do_patch() {
         fi
     else
         echo -e "${orange}${patchName}${reset}"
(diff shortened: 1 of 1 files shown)


ImportTaste:
Tested this on my (pre-existing) media-autobuild_suite. It went smoothly, no issues of any kind.

wiiaboo:
Can you split the changes by type/category? Like only = to == in a single commit, then indentation only, etc.?

build/media-suite_helper.sh
```diff
@@ -425,15 +425,16 @@ do_extract() {
...
-        { [[ $build32 = "yes" ]] && { [[ ! -f "$dirName"/build_successful32bit ]] ||
-            [[ -n "$flavor" && ! -f "$dirName/build_successful32bit_${flavor}" ]]; } ||
-          [[ $build64 = "yes" ]] && { [[ ! -f "$dirName"/build_successful64bit ]] ||
-            [[ -n "$flavor" && ! -f "$dirName/build_successful64bit_${flavor}" ]]; };
+    [[ -z $dirName ]] && dirName=$(guess_dirname "$archive")
+    if [[ $dirName != "." && -d $dirName ]] &&
+        {
+            [[ $build32 == "yes" ]] && { [[ ! -f "$dirName"/build_successful32bit ]] ||
+                [[ -n $flavor && ! -f "$dirName/build_successful32bit_${flavor}" ]]; } ||
+                [[ $build64 == "yes" ]] && { [[ ! -f "$dirName"/build_successful64bit ]] ||
```
wiiaboo: This doesn't seem right. Isn't this line supposed to have one less indent?

1480c1: I think it indented every time it hit `||`

1480c1:
wait, that's still wrong.

wiiaboo:
You could also duplicate the rm -rf $dirName if it would make the if condition clearer.

build/media-suite_helper.sh
```diff
@@ -429,19 +429,17 @@ do_extract() {
...
-    elif [[ -d $dirName ]]; then
+    if [[ $dirName != "." && -d $dirName ]]; then
+        if [[ $build32 == "yes" ]] &&
+            [[ ! -f "$dirName"/build_successful32bit || -n $flavor && ! -f "$dirName/build_successful32bit_${flavor}" ]]; then
+            rm -rf "$dirName"
+        elif [[ $build64 == "yes" ]] &&
+            [[ ! -f "$dirName"/build_successful64bit || -n $flavor && ! -f "$dirName/build_successful64bit_${flavor}" ]]; then
+            rm -rf "$dirName"
+        fi
+    elif [[ -d $dirName ]] || ! expr "$archive" : '.\+\(tar\(\.\(gz\|bz2\|xz\)\)\?\|7z\|zip\)$' > /dev/null; then
```
wiiaboo: The expr part does not do the same thing. If it's not an archive, it should not try to enter a non-existing dirName

PR #1363, opened 2019-08-24
vulkan-loader: unclamp

Helenerineium:


build/media-suite_compile.sh
```diff
@@ -1970,31 +1970,38 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
-        cd_safe Vulkan-Headers
-            _check=(vulkan/vulkan.h)
-            do_uninstall include/vulkan
-            do_cmakeinstall
+    if ! mpv_disabled vulkan; then
+
+        _check=(d3dkmthk.h d3dukmdt.h)
+        if ! files_exist "${_check[@]}" && do_vcs "https://github.com/DeadSix27/python_cross_compile_script.git"; then
+            do_uninstall "${_check[@]}"
+            do_install additional_headers/d3d{kmthk,ukmdt}.h include/
```
wiiaboo: Aren't these static? Why are we adding a VCS for this?

build/media-suite_compile.sh
```diff
@@ -1970,31 +1970,38 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
+            cd_safe Vulkan-Headers
+                _check=(vulkan/vulkan.h)
+                do_uninstall include/vulkan
+                do_cmakeinstall
+                do_checkIfExist
+                _check=(libvulkan.a vulkan.pc)
+            cd_safe "$LOCALBUILDDIR/$(get_first_subdir)"
             do_cmake -DBUILD_TESTS=no -DCMAKE_SYSTEM_NAME=Windows  \
-            -DCMAKE_ASM-ATT_COMPILER="$(command -v nasm.exe)" -DVULKAN_HEADERS_INSTALL_DIR="${LOCALDESTDIR}" \
+            -DCMAKE_ASM_COMPILER="$(command -v nasm.exe)" -DVULKAN_HEADERS_INSTALL_DIR="${LOCALDESTDIR}" \
```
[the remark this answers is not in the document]
Helenerineium: We patch the `if(WIN32)` to `if(MSVC)`, so the answer is yes.

build/media-suite_compile.sh
Helenerineium: Now using curl to download the headers, is that OK?

wiiaboo: Use do_wget instead. See other cases where it's just downloading files in the scripts.

build/media-suite_compile.sh
```diff
@@ -1992,30 +1992,30 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
         do_uninstall "${_check[@]}"
-        do_patch "$_shinchiro_patches/vulkan-0001-cross-compile-static-linking-hacks.patch"
+        do_patch "$_DeadSix27/patches/vulkan/0001-fix-cross-compiling.patch"
         create_build_dir
         log dependencies /usr/bin/python3 ../scripts/update_deps.py --no-build
         cd_safe Vulkan-Headers
-            _check=(vulkan/vulkan.h)
-            do_uninstall include/vulkan
+            _check=(vulkan/vulkan.h d3d{kmthk,ukmdt}.h)
+            do_uninstall include/vulkan d3d{kmthk,ukmdt}.h
```
wiiaboo: No need to keep uninstalling them, these types of headers don't get updated that often.

build/media-suite_compile.sh
```diff
@@ -1992,30 +1992,30 @@ if [[ $mpv != "n" ]] && pc_exists libavcodec libavformat libswscale libavfilter;
...
             do_cmakeinstall
+            do_wget -c -r -q "$_DeadSix27/additional_headers/d3dkmthk.h"
+            do_wget -c -r -q "$_DeadSix27/additional_headers/d3dukmdt.h"
+            do_install d3d{kmthk,ukmdt}.h include/
             do_checkIfExist
             _check=(libvulkan.a vulkan.pc)
         cd_safe "$LOCALBUILDDIR/$(get_first_subdir)"
-        CFLAGS+=" -D_WIN32_WINNT=0x0600 -D__STDC_FORMAT_MACROS" \
-            CPPFLAGS+=" -D_WIN32_WINNT=0x0600 -D__STDC_FORMAT_MACROS" \
-            CXXFLAGS+=" -D__USE_MINGW_ANSI_STDIO -D__STDC_FORMAT_MACROS -fpermissive -D_WIN32_WINNT=0x0600" \
```
wiiaboo: These Cflags aren't needed anymore?

Helenerineium: The default of _WIN32_WINNT is [0x0601](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-headers-git/PKGBUILD#L58)
__USE_MINGW_ANSI_STDIO is already defined in our batch file
__STDC_FORMAT_MACROS is [not needed](https://en.cppreference.com/w/cpp/types/integer#Notes) if you have c++11
The CMakeLists sets `-std=c++11` [here](https://github.com/KhronosGroup/Vulkan-Loader/blob/160e66b55a385eec961187b6e422417541c25608/CMakeLists.txt#L150)
`-fpermissive` is not needed and shouldn't be used IMO

Issue #1364, opened 2019-08-24
[mingw64] [ffmpeg with libsvtav1] libsvtav1 and libaom seems to conflict with each other


moisesmcardona:
I'm not sure why I was not seeing this make error before, but now that I'm disabling and testing different options in order to successfully use SVT-AV1 in ffmpeg with most plugins, I came to this issue:

```
Likely error:
D:/repos/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/repos/media-autobuild_suite/local64/lib/libSvtAv1Enc.a(convolve_avx2.c.obj):convolve_avx2.c:(.text+0xc90): multiple definition of `av1_build_compound_diffwtd_mask_d16_avx2'; D:/repos/media-autobuild_suite/local64/lib/libaom.a(reconinter_avx2.c.o):reconinter_avx2.c:(.text+0x350): first defined here

D:/repos/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/repos/media-autobuild_suite/local64/lib/libSvtAv1Enc.a(convolve_avx2.c.obj):convolve_avx2.c:(.text+0x1cc0): multiple definition of `av1_wedge_sse_from_residuals_avx2'; D:/repos/media-autobuild_suite/local64/lib/libaom.a(wedge_utils_avx2.c.o):wedge_utils_avx2.c:(.text+0x0): first defined here

D:/repos/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/repos/media-autobuild_suite/local64/lib/libSvtAv1Enc.a(convolve_avx2.c.obj):convolve_avx2.c:(.text+0x1da0): multiple definition of `aom_subtract_block_avx2'; D:/repos/media-autobuild_suite/local64/lib/libaom.a(subtract_avx2.c.o):subtract_avx2.c:(.text+0x0): first defined here

[...]

PR #1365, opened 2019-08-25
Work around for issues if either libsvthevc and/or libsvtav1 are enabled

moisesmcardona:
If `libsvthevc` and `libsvtav1` are enabled, `libsvtav1` will be disabled because the patch for `libsvthevc` has been enabled and `libsvtav1` currently does not run if `libsvthevc` is enabled. It crashes ffmpeg.

On the other hand, if `libsvthevc` is disabled and `libsvtav1` is enabled, the `libsvtav1` patch will be applied. 

Lastly, it will disable `libaom` and `libopencore-amrwb` if they are enabled and `libsvtav1` is enabled, because `libsvtav1` cannot run due to code conflicts in them, which, while make ffmpeg compile successfully, `libsvtav1` will not run and ffmpeg will crash.

See the following issues for more information:
https://github.com/m-ab-s/media-autobuild_suite/issues/1364
https://github.com/OpenVisualCloud/SVT-AV1/issues/567


build/media-suite_compile.sh
```diff
@@ -1750,14 +1750,34 @@ if [[ $ffmpeg != "no" ]]; then
                 do_removeOption --enable-libsvthevc
         fi
 
+        # -- SVT-AV1 patches for ffmpeg --
+
+        # See issue https://github.com/OpenVisualCloud/SVT-AV1/issues/567 for the reasons behind the changes below.
+        # This is broken in the meantime. Either libsvthevc or libsvtav1 can be enabled, but not both at the same time.
+        #remove the do_removeOption line and uncomment the patch once the issue have been fixed.
         if enabled libsvthevc && enabled libsvtav1; then
```
1480c1: Can you move this check to before applying svt-hevc's patch?

1480c1:
To add on: After you are done with all changes and are sure there will be no more, you can do
``` bash
git fetch --all -pf
git rebase -i upstream/master # Mark all/small commits as fixup into logical commits
git diff origin/libsvtav1-temporary-fix # to check to make sure no additional changes slipped through
git push -f
```
Doing this can help keep the commit list and history clean plus it can make sure it's merge-able into master

https://help.github.com/en/articles/about-git-rebase
https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

wiiaboo:
People opening issues are not going to look at code for known issues.
Disable svt-av1 on the .bat meanwhile and both issues are dealt with for now.

moisesmcardona:
Technically, the SVT-AV1 partially works. I believe this PR will help those who wants to have a working SVT-AV1 regardless of the enabled libraries. Otherwise, they may run into the same issues as I did during these days.

Once the issues have been solved in upstream, we can revert it.

build/media-suite_compile.sh
moisesmcardona: done.

moisesmcardona:
To add, if a user wants to add `libsvtav1` they need to disable `libsvthevc`. Leaving both turned on will disable `libsvtav1`.

This PR is kind of tricky, because we may choose to do it the other way, disabling `libsvthevc` instead and leaving `libsvtav1` enabled. I chose the first because the second requires disabling more libraries.

ImportTaste:
I'm honestly not an expert like @wiiaboo and @1480c1, but it seems to me that if you're going to make a workaround, you should have the code alter as little of the original script as possible to make it easy to revert later without any possible merge conflict.

For example, you're removing this section of code, even though the check you're applying before it would make `enabled libsvthevc` return false anyway.
https://github.com/m-ab-s/media-autobuild_suite/blob/master/build/media-suite_compile.sh#L1753-L1756

If I'm wrong you can disregard me entirely, but I would redo this so it stands on its own as much as possible while not removing any of the existing code. If some of the existing code absolutely cannot run, maybe just comment it out?

moisesmcardona:
The code isn't removed. It is just commented, because at the current state, it doesn't make sense to apply the patch that should make both SVT-HEVC and SVT-AV1 work, since it fails anyway. SVT-HEVC will work, while SVT-AV1 will crash ffmpeg.

You're probably right in that the if/elif statement should have been kept the same. I'm doing a test...

ImportTaste:
This is more what I have in mind. I have not tested this.

https://github.com/ImportTaste/media-autobuild_suite/pull/1/commits/aeed2a14bc68b8949a75f2181a79869e1a7f4b62

It is entirely self-contained, because it occurs before any other checks, it will prevent the checks that follow from resolving how they normally would. (e.g. `0001-Add-ability-for-ffmpeg-to-run-svt-av1-with-svt-hevc.patch` won't be applied because `enabled libsvtav1` will no longer return true because this codeblock ran `do_removeOption --enable-libsvtav1`)

So when the issue is fixed, all one needs to do is remove this codeblock, rather than removing scattered changes and uncommenting code.

moisesmcardona:
See my above commit. Yours is essentially the same without the libsvthevc and libsvtavc1 patch commented like mine, and nested, as well as with the print statements. If we want to add those print statements, I could add it too. You just basically nested the checks within the current if/elif.

ImportTaste:
> See my above commit. Yours is essentially the same without the libsvthevc and libsvtavc1 patch commented like mine, and nested, as well as with the print statements. If we want to add those print statements, I could add it too. You just basically nested the checks within the current if/elif.

This is an entirely new code block that runs before all the other current code.

moisesmcardona:
I only see that you relocated and converted the ifs into if/elifs and added
some print statements... But anyway, if it accomplishes the same, that's
the goal...

ImportTaste:
> I only see that you relocated and converted the ifs into if/elifs and added some print statements... But anyway, if it accomplishes the same, that's the goal...

This is all just one solid chunk and it doesn't have to comment out any existing code.

moisesmcardona:
In that case, feel free to open a PR and let the devs choose which variants
they prefer. This is just temporary, anyway, altough we don't know if, or
when, the problems will get fixed on upstream.

ImportTaste:
@1480c1 I don't want to bother wiiaboo since he seems like he's pretty busy (and grumpy) so I'm curious which implementation you think would be better.

Mine runs redundant code at the beginning to keep it all clustered together since I don't think the upstream is going to be bugged like this for too long, but I could be wrong. https://github.com/ImportTaste/media-autobuild_suite/commit/aeed2a14bc68b8949a75f2181a79869e1a7f4b62

moisesmcardona's implementation comments out code and has the temporary code run in about the same area as the original code. (His also lacks the printed alerts but those could easily be carried over.)

If mine is preferred I'll do a PR.

1480c1:
Just create a pr, it makes it easier to see

PR #1370, opened 2019-08-27
bat: Revert permission changes

1480c1:


wiiaboo:
For trivial commits that you don't expect there to be discussion or controversy there's no need for PRs. Master isn't protected.

PR #1372, opened 2019-08-28
SVT-AV1: Add patch to prevent library issues when compiling ffmpeg

moisespr123:
There's work being done to solve the library issues we're having when compiling ffmpeg with libsvtav1 in their PR https://github.com/OpenVisualCloud/SVT-AV1/pull/573

I added this patch and removed he check for `libopencore-amrwb` as the issue with this library has been fixed.

Still need to be fixed are issues with `libaom` and `libsvthevc`.

Opening this as a draft until the patch can be fully added and the `libsvtav1` checks can be removed.


ImportTaste:
A nested if check shouldn't be necessary anymore, so this can be shortened to
```
        if enabled libsvtav1 && enabled libaom; then
            do_print_progress "Until SVT-AV1 issues a fix, libaom must be disabled while libsvtav1 is enabled."
            do_removeOption --enable-libaom
        fi
```

PR #1373, opened 2019-09-01
rav1e: Add flags to do_rust to address rav1e issues on Windows on CPU's with AVX2

moisesmcardona:
Recently, there's an issue with rav1e when using it on CPU's with AVX2.

This PR adds some flags that fixes the issue, as suggested on this issue that I also opened over there: https://github.com/xiph/rav1e/issues/1618

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -971,7 +971,7 @@ if { [[ $rav1e = y ]] || enabled librav1e; } &&
 
     # standalone binary
     if [[ $rav1e = y || $standalone = y ]]; then
-        do_rust
+        do_rust --no-default-features --features=binaries
         do_install "target/$CARCH-pc-windows-gnu/release/rav1e.exe" bin-video/
     fi
 
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1957,7 +1957,9 @@ do_rust() {
     extra_script pre rust
     log "build" "$RUSTUP_HOME/bin/cargo.exe" build --release \
         --target="$CARCH"-pc-windows-gnu \
-        --jobs="$cpuCount" "$@"
+        --jobs="$cpuCount" "$@" \
+		--no-default-features \
+		--features=binaries
     extra_script post rust
 }
 
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1957,9 +1957,7 @@ do_rust() {
     extra_script pre rust
     log "build" "$RUSTUP_HOME/bin/cargo.exe" build --release \
         --target="$CARCH"-pc-windows-gnu \
-        --jobs="$cpuCount" "$@" \
-		--no-default-features \
-		--features=binaries
+        --jobs="$cpuCount" "$@"
     extra_script post rust
 }
 


1480c1:
Can you make sure that this doesn't affect other rust projects? (ripgrep, dssim, and rav1e are the only rust projects atm I think) If unsure, just apply it to rav1e https://github.com/m-ab-s/media-autobuild_suite/blob/d73be168704e92bf47a9c4796a504570fd3d49ab/build/media-suite_compile.sh#L974

moisesmcardona:
I'll verify tomorrow. Going to sleep now.

1480c1:
Good night

jamrial:
This isn't a fix. You're disabling features (assembly among them) to workaround the problem.

1480c1:
Is there/Do you recommend some other recommended way to work around the problem temporarily?

jamrial:
Not really, this is probably the only workaround until the issue is fixed upstream. But it will effectively generate crippled builds even for people unaffected by the crash (pre AVX2 cpus).

1480c1:
Okay, I was thinking about trying to detect avx2 or not and then disable the features, but that would also harm distributing builds

Is there a way to disable avx2 only? (I'm assuming that avx2 functions are causing the problems) I'm not very familiar with cargo build's commands and features

jamrial:
I don't think so, but don't quote me on this, i'm not familiar with cargo either.

It may be a good idea to ask in the above rav1e ticket for suggestions, fwiw. But hopefully it will be fixed soon, in any case.

ImportTaste:
It probably isn't going to take long for this to be fixed upstream if https://github.com/xiph/rav1e/pull/1621 is any indicator. Unless that PR ends up taking a while (and as of writing this it hasn't even been 24hrs yet), a temporary fix shouldn't be necessary.

ImportTaste:
@1480c1 A thought comes to mind. Instead of trying to stay on top of every upstream issue with up-to-the-minute temporary fixes, would it be possible to extend the Custom Patches feature to allow the override of a media-suite_helper.sh function?

For instance, `do_rust` could run `do_rust_override` instead if detected (with a check to make sure do_rust_override won't try and run itself if a user doesn't remove that line in their copy-pasted version). Or, something similar to `touch do_not_reconfigure` could be used.

Aside from giving users greater control, it would allow you or wiiaboo to give them an override to try where normally you'd have to have them modify one of the scripts or use one from a forked branch. There would also be less pressure to push a temporary fix if it hasn't been broken upstream for very long, users could just use the override supplied in a given issue.

The only problem I see with this approach is it might lead to cases where someone forgets about an override, but that problem exists with other custom patches too.

moisesmcardona:
I didn't notice that arguments could be passed to `do_rust()`. My bad. I moved them to the rav1e portion of the code.

That said, I'd like to know how does this repo handles patches like this, that while rav1e builds, it doesn't work, and some temporary patches need to be done.

I don't want to open unnecessary or unwelcomed PR's, but there are times that measures need to be done in order to have a successfully built and working product... Be it permanent or temporary.

wiiaboo:
Well, rav1e without asm probably couldn't be considered as "working". I'd prefer clamping to a previous working commit than the suite producing impossibly slow encoders.

wiiaboo:
> For instance, `do_rust` could run `do_rust_override` instead if detected (with a check to make sure do_rust_override won't try and run itself if a user doesn't remove that line in their copy-pasted version). Or, something similar to `touch do_not_reconfigure` could be used.
> 
> Aside from giving users greater control, it would allow you or wiiaboo to give them an override to try where normally you'd have to have them modify one of the scripts or use one from a forked branch. There would also be less pressure to push a temporary fix if it hasn't been broken upstream for very long, users could just use the override supplied in a given issue.

That's already implemented as custom scripts. What's the point of creating yet another way of overriding what the suite does?

wiiaboo:
I'd rather this be a temporary workaround applied on a case-by-case. Leave an open issue while it's not fixed upstream or something.

ImportTaste:
@wiiaboo I read the documentation and I couldn't find a way to override do_rust's compiling flags. In this instance, how would you make the suite compile AVX2 with the `--no-default-features` and `--features=binaries` flags?

ImportTaste:
>That's already implemented as custom scripts. What's the point of creating yet another way of overriding what the suite does?

I just noticed the latest commits @wiiaboo, I think you must've misunderstood me. What you're doing is pretty close to what I was suggesting. And it's very much appreciated.

wiiaboo:
```bash
create_extra_skeleton rav1e
sed -i '/_pre_rust/a rust_extras+=(--no-default-features --features=binaries)' /build/rav1e_extra.sh
```

ImportTaste:
Thanks!

PR #1374, opened 2019-09-01
Rav1e fix

moisesmcardona:
Temporary rav1e workaround, until the issue is fixed in rav1e.


moisesmcardona:
Oops, my bad. Wasn't meaning to open it in this repo...

PR #1384, opened 2019-09-16
create_build_dir: Add nocd and print options

1480c1:
Add nocd and print option accessible using `create_build_dir -c` and `-p` options in order to be able to create a directory without necessarily cding into it and print a full path to pass to other tools (meson and cmake)

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1379,16 +1379,35 @@ log() {
 }
 
 create_build_dir() {
-    local build_dir="build${1:+-$1}-$bits"
-    if [[ ! -f "$(get_first_subdir)/do_not_clean" ]]; then
-        if [[ "$(basename "$(pwd)")" == "$build_dir" ]]; then
-            rm -rf ./* && cd_safe ".."
-        elif [[ -d $build_dir ]] && ! rm -rf ./"$build_dir"; then
-            cd_safe "$build_dir" && rm -rf ./* && cd_safe ".."
-        fi
+    local print_build_dir=false nocd=${nocd:-false} norm=false build_root build_dir getoptopt OPTARG OPTIND
+    build_root="$LOCALBUILDDIR/$(get_first_subdir)"
+    while getopts ":pcrC:" getoptopt; do
+        case $getoptopt in
+        p) print_build_dir=true ;;
+        c) nocd=true ;;
+        r) norm=true ;;
+        C) build_root="$OPTARG" ;;
+        \?)
+            echo "Invalid Option: -$OPTARG" 1>&2
+            return 1
+            ;;
+        :)
+            echo "Invalid option: $OPTARG requires an argument" 1>&2
+            return 1
+            ;;
+        esac
+    done
+    shift $((OPTIND - 1))
+
+    build_dir="$build_root/build${1:+-$1}-$bits"
+
+    if [[ -d $build_dir && ! -f $LOCALBUILDDIR/$(get_first_subdir)/do_not_clean ]]; then
+        $norm || rm -rf "$build_dir" ||
+            (cd_safe "$build_dir" && rm -rf ./*)
     fi
-    [[ ! -d $build_dir ]] && mkdir "$build_dir"
-    cd_safe "$build_dir"
+    [[ ! -d $build_dir ]] && mkdir -p "$build_dir"
+    $nocd || cd_safe "$build_dir"
+    $print_build_dir && printf '%s\n' "$build_dir"
 }
 
 get_external_opts() {


wiiaboo:
Is that second functionality being used somewhere?

1480c1:
Planning to use, but separated the commit for easier review

build/media-suite_helper.sh
```diff
@@ -1387,7 +1404,18 @@ create_build_dir() {
...
     fi
     [[ ! -d $build_dir ]] && mkdir "$build_dir"
-    cd_safe "$build_dir"
+    $nocd || cd_safe "$build_dir"
+    if $print_build_dir; then
+        printf '%s' "$(
+            cd_safe "$(dirname "$build_dir")"
+            pwd -P
+        )"
+        if [[ $build_dir == "$(basename "$PWD")" ]]; then
```
wiiaboo: What does this if-else do? Only the above printf is showing up for me.

1480c1: That's to filter out when running in the build directory (to prevent build-64bit/build-64bit)

wiiaboo: That doesn't happen because of the code above.
And actually, the code never works unless it's run from /build because get_first_subdir only returns the dirname, not the full path. Both the do_not_clean and do_not_reconfigure checks don't work.

1480c1: Hmm, it seems I need to relook at all of the code

build/media-suite_helper.sh
```diff
@@ -1378,16 +1378,36 @@ log() {
...
+        esac
+    done
+    shift $((OPTIND - 1))
+
+    build_dir="$build_root/build${1:+-$1}-$bits"
+
+    if [[ -d $build_dir ]]; then
+        [[ ! -f $LOCALBUILDDIR/$(get_first_subdir)/do_not_clean ]] &&
+            ! rm -rf "$build_dir" &&
+            (cd_safe "$build_dir" && rm -rf ./*) # Should theoretically not get here
```
1480c1:
```suggestion
            (cd_safe "$build_dir" && rm -rf ./*)
```

build/media-suite_helper.sh
```diff
@@ -1378,16 +1378,36 @@ log() {
...
-        elif [[ -d $build_dir ]] && ! rm -rf ./"$build_dir"; then
-            cd_safe "$build_dir" && rm -rf ./* && cd_safe ".."
-        fi
+    local print_build_dir=false nocd=${nocd:-false} norm=false build_root repo_dir build_dir getoptopt OPTARG OPTIND
+    build_root="$LOCALBUILDDIR/$(get_first_subdir)"
+    while getopts ":pcrC:" getoptopt; do
+        case $getoptopt in
+        p) print_build_dir=true ;;
+        c) nocd=true ;;
+        r) norm=true ;;
```
wiiaboo: Is this option supposed to be doing something?

1480c1: I feel like it was supposed to be an alternate to `do_not_clean` but for use with printing a build dir without cleaning it, but I guess I forgot to actually implement it

1480c1: Added in line 1406

build/media-suite_helper.sh
```diff
@@ -1379,16 +1379,37 @@ log() {
...
+            return 1
+            ;;
+        esac
+    done
+    shift $((OPTIND - 1))
+
+    build_dir="$build_root/build${1:+-$1}-$bits"
+
+    if [[ -d $build_dir ]]; then
+        [[ ! -f $LOCALBUILDDIR/$(get_first_subdir)/do_not_clean ]] &&
```
wiiaboo: Shouldn't it be an OR instead?

1480c1: No since it should first check if `do_not_clean` does not exist, then if norm is false, then try to remove the files

1480c1: I think

wiiaboo: But you need do_not_clean to be there so the option works. If the file doesn't exist the option does nothing. Isn't that redudant?

wiiaboo: If you always need to do two things to do a specific action, but if you forget one of them, then it should be a single thing anyway.

1480c1: I'll check the logic again, I'm confusing myself

1480c1: Switched around the logic

wiiaboo: Looks more or less the same. Don't you still need do_not_clean for the option to do anything?

1480c1: No, as if `do_not_clean` exists, create_build_dir should not delete any files and not do anything, so it should skip the whole block if it exists

wiiaboo: Right.

PR #1386, opened 2019-09-17
rust: add sccache

1480c1:
Add sccache (ccache but for rustc etc)

Combine jq checks since they both check for jq.

Technically speaking, we could try to replace ccache with sccache, problem is that we can't disable sccache like ccache.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -169,6 +169,15 @@ if [[ $ripgrep = y || $rav1e = y || $dssim = y ]]; then
     log rustup_update "$RUSTUP_HOME/bin/rustup.exe" update
     log set_default_toolchain "$RUSTUP_HOME/bin/rustup.exe" default \
         "stable-$CARCH-pc-windows-gnu"
+
+    _check=(bin/sccache.exe)
+    if do_vcs "https://github.com/mozilla/sccache.git"; then
+        do_uninstall "${_check[@]}"
+        do_rust
+        do_install "target/$CARCH-pc-windows-gnu/release/sccache.exe" bin/
+        do_checkIfExist
+        export RUSTC_WRAPPER=sccache
+    fi
 fi
 
 _check=(bin-global/rg.exe)
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1273,6 +1273,7 @@ do_mesoninstall() {
 
 do_rust() {
     log "rust.update" "$RUSTUP_HOME/bin/cargo.exe" update
+    command -v sccache > /dev/null 2>&1 && export RUSTC_WRAPPER=sccache
     # use this array to pass additional parameters to cargo
     local rust_extras=()
     extra_script pre rust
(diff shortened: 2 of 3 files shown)


wiiaboo:
I actually don't know if we still need oniguruma. We thought the jq issue was caused by not compiling our own, but it turned out to be broken by out-of-root builds. Probably try to revert again to using mingw's.

build/media-suite_compile.sh
```diff
@@ -161,14 +161,23 @@ if [[ $ripgrep = y || $rav1e = y || $dssim = y ]]; then
         add_to_remove
         cd_safe "$LOCALBUILDDIR"
     fi
-    if ! [[ $(rustup toolchain list) =~ stable-$CARCH-pc-windows-gnu ]]; then
+    if ! rustup toolchain list | grep -q "stable-$CARCH-pc-windows-gnu"; then
```
wiiaboo: There any special reason for this? The bash regex check is lighter on resources than using grep.

1480c1:
Okay

build/media-suite_compile.sh
1480c1: Mainly for readability since it's easier to spot a grep than `=~`

I can revert if it's really unnecessary

build/media-suite_helper.sh
```diff
@@ -1273,6 +1273,8 @@ do_mesoninstall() {
 
 do_rust() {
     log "rust.update" "$RUSTUP_HOME/bin/cargo.exe" update
+    export SCCACHE_DIR=~/.sccache
```
wiiaboo: I'd rather `$HOME/.sccache`. I don't know if msys2 always expands the `~` correctly.

1480c1:
```suggestion
    export SCCACHE_DIR=$HOME/.sccache
```

media-autobuild_suite.bat
```diff
@@ -1718,6 +1718,8 @@ goto :EOF
     echo.export DXSDK_DIR ACLOCAL_PATH PKG_CONFIG PKG_CONFIG_PATH CPPFLAGS CFLAGS CXXFLAGS LDFLAGS MSYSTEM
     echo.
     echo.export CARGO_HOME="/opt/cargo" RUSTUP_HOME="/opt/cargo"
+    echo.export SCCACHE_DIR=~/.sccache
```
```suggestion
    echo.export SCCACHE_DIR="$HOME/.sccache"
```

media-autobuild_suite.bat
```diff
@@ -1718,6 +1718,8 @@ goto :EOF
...
+    echo.export CCACHE_DIR=~/.ccache
```
```suggestion
    echo.export CCACHE_DIR="$HOME/.ccache"
```

build/media-suite_helper.sh
1480c1:
```suggestion
    export SCCACHE_DIR="$HOME/.sccache"
```

build/media-suite_helper.sh
```diff
@@ -1273,6 +1273,8 @@ do_mesoninstall() {
 
 do_rust() {
     log "rust.update" "$RUSTUP_HOME/bin/cargo.exe" update
+    export SCCACHE_DIR=$HOME/.sccache
```
1480c1:
```suggestion
    export SCCACHE_DIR="$HOME/.sccache"
```

build/media-suite_compile.sh
```suggestion
    if ! [[ $(rustup toolchain list) =~ stable-$CARCH-pc-windows-gnu ]]; then
```

1480c1:
Any more changes needed before I rebase?

1480c1:
> I actually don't know if we still need oniguruma. We thought the jq issue was caused by not compiling our own, but it turned out to be broken by out-of-root builds. Probably try to revert again to using mingw's.

Should I strip the jq stuff from this pr then?

1480c1:
Moved the jq stuff to another branch for now

PR #1387, opened 2019-09-17
do_configure: Move prefix to default

1480c1:
Move prefix option from separateconf to do_configure.

Removed the build,host,target options since those _should_ be auto detected (but some configure scripts don't accept them)

Doesn't seem to break anything yet


wiiaboo:
The ffmpeg ones can be merged (including x264 and cyan). If one works the other will also.
Did you test the ones with only --host or --target to see if anything broke? Sometimes the configures complain if there's not at least one of them.

1480c1:
Right now testing for the lack of host,build,target

wiiaboo:
Merged in d2f011047575d618f47cf1ae0f4c752a1f69ec0a

PR #1397, opened 2019-09-22
vulkan-loader: use do_cmakeinstall

Helenerineium:


1480c1:
Do the libs etc get installed to the proper folder? 

Right now running with this applied, but vulkan-loader failed due to not finding vulkan-headers. Seems update_deps.py did not run or something, rerunning

1480c1:
Seems to work

Issue #734, opened 2018-01-31
xvc support


Selur:
Would be nice if https://github.com/divideon/xvc could be added.

wiiaboo:
Sounds like a shit codec if it's "based" on MPEG stuff and changing the codec if you get sued.
Why would anyone want this?

Selur:
To try whether it is good or not.

wiiaboo:
Well, if you just wanna try it, just run mintty and
```bash
cd /build
source media-suite_helper.sh
git clone https://github.com/divideon/xvc.git
cd xvc
do_cmake -DBUILD_TESTS=off
ninja
```
and there you go.
I think I'll pass on adding a lawsuit liability and waste of time codec no one will ever use.

LigH-de:
PS: Was the patch now included?

```
┌ xvc git  ............................................... [Updates found]
├ Running uninstall...
15.patch
        Patch could not be applied with `git am`. Continuing without patching.
```

1480c1:
Yep https://github.com/m-ab-s/media-autobuild_suite/commit/1c6b35eefb6029fa0116772da5cdf5dda2dfbc27

PR #1398, opened 2019-09-22
xvc: Add xvc

1480c1:
Add xvc: https://github.com/divideon/xvc
https://xvc.io

Claims to be a HEVC and AV1 competitor in terms of performance.

Seems to be interesting enough, however, since it doesn't have an ffmpeg plugin or others, I feel that it may be limited in use.

Related doom9: https://forum.doom9.org/showthread.php?t=175110

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ For information about the compiler environment see the wiki, there you also have
     - x264 (8 and 10 bit, with l-smash [mp4 output], lavf and ffms2) (git)
     - x265 (8, 10 and 12 bit) (hg)
         - includes svt-hevc encoder if enabled
+    - xvc (git) (unsupported)
     - xvid (1.3.5)
 
 --------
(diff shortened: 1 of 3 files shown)


wiiaboo:
I was pretty sure I'd refused this encoder before because the licensing [sounds stupid](https://github.com/m-ab-s/media-autobuild_suite/issues/734). [nev](https://forum.doom9.org/showthread.php?p=1828091#post1828091) also feels like this makes for a codec with no guarantee of future or retrocompatibility, meaning there's no point in using it other than for curiosity. Big platforms will never use it, archival will never use it, so 🤷

You can still add it, though, since this is a PR and you did the work.

Helenerineium:
Unofficial ffmpeg patch can be found here: https://github.com/perher/FFmpeg/commits/release/4.0-xvc
Requires some work if you want to apply it on master, but not too much.
You can also add it if interested.

1480c1:
I've added a note about low-prio on issues similar to mplayer and ffmbc

PR #1402, opened 2019-09-23
do_rust: Unset RUSTC_WRAPPER if command -v sccache fails

### LINKED ISSUE #1399: failed sccache build
┌ sccache git  ........................................ [Recently updated]
├ Running rust.update...
├ Running rust.build...
Likely error (tail of the failed operation logfile):
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
error: failed to run `rustc` to learn about target-specific information

Caused by:
  could not execute process `sccache rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro` (never executed)

Caused by:
  Det går inte att hitta filen. (os error 2)
rust.build failed. Check D:/media-ab_suite/build/sccache-git/ab-suite.rust.build.log
  Creating diagnostics file...

All relevant logs have been anonymously uploaded to https://0x0.st/ztt8.zip
Copy and paste [logs.zip](https://0x0.st/ztt8.zip) in the GitHub issue.

1480c1:
Should Fix #1399 and #1403

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -173,8 +173,8 @@ if [[ $ripgrep = y || $rav1e = y || $dssim = y ]]; then
 
     _check=(bin/sccache.exe)
     if do_vcs "https://github.com/mozilla/sccache.git"; then
-        do_uninstall "${_check[@]}"
         do_rust
+        sccache --stop-server >/dev/null 2>&1 || true
         do_install "target/$CARCH-pc-windows-gnu/release/sccache.exe" bin/
         do_checkIfExist
         export RUSTC_WRAPPER=sccache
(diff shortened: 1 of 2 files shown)


build/media-suite_compile.sh
```diff
@@ -173,8 +173,8 @@ if [[ $ripgrep = y || $rav1e = y || $dssim = y ]]; then
 
     _check=(bin/sccache.exe)
     if do_vcs "https://github.com/mozilla/sccache.git"; then
-        do_uninstall "${_check[@]}"
         do_rust
+        sccache --stop-server >/dev/null 2>&1 || false
```
wiiaboo: Shouldn't we check if sccache exists first?

```suggestion
        sccache --stop-server >/dev/null 2>&1 || true
```

That's what the to dev nulll and 2>&1 are for, just ignore if it fails

But I did make the mistake of putting false there

PR #1407, opened 2019-09-26
Batch: Improve a couple things

1480c1:
Contains improvements for noMintty through use script.exe, now doesn't require the hackish bash.ps1

Includes some style changes towards using `::` as comments and replace them with proper `rem` comments since technically speaking `::` is not actually a comment but a label with `:` as the name

Tested with noMintty=1 and =0

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1463,39 +1463,30 @@ rem installbase
 if exist "%instdir%\%msys2%\etc\pac-base.pk" del "%instdir%\%msys2%\etc\pac-base.pk"
 for %%i in (%msyspackages%) do echo.%%i>>%instdir%\%msys2%\etc\pac-base.pk
 
-if exist %instdir%\%msys2%\usr\bin\make.exe GOTO sethgBat
-echo.-------------------------------------------------------------------------------
-echo.install msys2 base system
-echo.-------------------------------------------------------------------------------
-if exist %build%\install_base_failed del %build%\install_base_failed
-(
-    echo.echo -ne "\033]0;install base system\007"
-    echo.msysbasesystem="$(cat /etc/pac-base.pk | tr '\n\r' '  ')"
-    echo.[[ "$(uname)" = *6.1* ]] ^&^& nargs="-n 4"
-    echo.echo $msysbasesystem ^| xargs $nargs pacman -Sw --noconfirm --ask=20 --needed
-    echo.echo $msysbasesystem ^| xargs $nargs pacman -S --noconfirm --ask=20 --needed
-    echo.echo $msysbasesystem ^| xargs $nargs pacman -D --asexplicit
-    echo.sleep ^3
-    echo.exit
-)>%build%\pacman.sh
-call :runBash pacman.log "%build%\pacman.sh"
-del %build%\pacman.sh
-
-for %%i in (%instdir%\%msys2%\usr\ssl\cert.pem) do (
-    if %%~zi==0 (
-        (
-            echo.update-ca-trust
-            echo.sleep ^3
-            echo.exit
-        )>%build%\cert.sh
-        call :runBash cert.log "%build%\cert.sh"
-        del %build%\cert.sh
-    )
+if not exist %instdir%\%msys2%\usr\bin\make.exe (
+    echo.-------------------------------------------------------------------------------
+    echo.install msys2 base system
+    echo.-------------------------------------------------------------------------------
+    if exist %build%\install_base_failed del %build%\install_base_failed
+    title install base system
+    (
+        echo.echo -ne "\033]0;install base system\007"
+        echo.msysbasesystem="$(cat /etc/pac-base.pk | tr '\n\r' '  ')"
+        echo.[[ "$(uname)" = *6.1* ]] ^&^& nargs="-n 4"
+        echo.echo $msysbasesystem ^| xargs $nargs pacman -Sw --noconfirm --ask=20 --needed
+        echo.echo $msysbasesystem ^| xargs $nargs pacman -S --noconfirm --ask=20 --needed
+        echo.echo $msysbasesystem ^| xargs $nargs pacman -D --asexplicit
+        echo.sleep ^3
+        echo.exit
+    )>%build%\pacman.sh
+    call :runBash pacman.log /build/pacman.sh
+    del %build%\pacman.sh
 )
 
-:sethgBat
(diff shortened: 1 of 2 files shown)


media-autobuild_suite.bat
```diff
@@ -1581,11 +1581,10 @@ if exist %build%\compilation_failed del %build%\compilation_failed
 if exist %build%\fail_comp del %build%\compilation_failed
 
 REM Test mklink availability
-set symlinkSupport=""
```
wiiaboo: Didn't even notice this was mistyped, heh.

1480c1: same, I only noticed it when I realized it's better to just go ahead and use the MSYS variable

wiiaboo:
Commit f2999b0f6f42bc175b9ae9d1cb1183d196390a92 looks like it's empty. How'd you manage to do that with git? I thought you couldn't do empty commits.

1480c1:
Possibly I fixed up a change that essentially nulled it, but I didn't rebase again.

Git does not seem to be smart enough to remove it the first time around

1480c1:
Will rebase to remove the empty commit

media-autobuild_suite.bat
```diff
@@ -1337,49 +1337,39 @@ if not exist %instdir%\mintty.lnk (
     echo -------------------------------------------------------------------------------
     echo.- make a first run
     echo -------------------------------------------------------------------------------
-    call :runBash firstrun.log "-c exit"
+    call :runBash firstrun.log exit
+
+    sed -i "s/#Color/Color/;s/^^IgnorePkg.*/#&/" %instdir%\%msys2%\etc\pacman.conf
```
wiiaboo: Is the color thing needed? You could always add `--color <when>` to select pacman commands if you really want to add colors.

media-autobuild_suite.bat
```diff
@@ -1337,49 +1337,39 @@ if not exist %instdir%\mintty.lnk (
...
-    call :runBash firstUpdate.log "%build%\firstUpdate.sh"
-    del %build%\firstUpdate.sh
+    title first msys2 update
+    call :runBash firstUpdate.log pacman --noconfirm -Sy --asdeps pacman-mirrors
 
     echo.-------------------------------------------------------------------------------
     echo.critical updates
     echo.-------------------------------------------------------------------------------
-    %bash% -lc "pacman -S --needed --ask=20 --noconfirm --asdeps bash pacman msys2-runtime"
+    pacman -S --needed --ask=20 --noconfirm --asdeps bash pacman msys2-runtime
```
wiiaboo: Isn't the `call :runBash` missing here? Does raw calling pacman work from batch?

media-autobuild_suite.bat
```diff
@@ -1373,6 +1373,40 @@ if not exist %instdir%\mintty.lnk (
     del %build%\setlink.vbs
 )
 
+rem createFolders
```
wiiaboo: This commit looks like new code is being added instead of being moved/refactored somewhere else. You can join the commits removing the old code and adding the new one to make it more atomic.

media-autobuild_suite.bat
1480c1: It's not really, just a convenience thing when using the mintty lnk outside of the script

media-autobuild_suite.bat
1480c1: After extensive testing on using pacman directly from batch, cmd, and powershell, pacman seems to work fine outside of bash.

A long time ago, I did have some problems with pacman complaining about missing commands, but I think that was because I called pacman by full path without having the msys2 folders in PATH

we could use runBash with this, but since it didn't have a log file named nor using mintty previously, I didn't. Plus since it's updating bash, pacman, and msys2-runtime, I think it will ask to close without calling exit and Idk if there's a way to easily make it close without user interaction

media-autobuild_suite.bat
1480c1: I honestly did have it like that, but, I had too many changes, that I reset to upstream and did commit -p

media-autobuild_suite.bat
wiiaboo: I usually add it myself, but sure, you can leave it. The logs are pruned of color characters anyway.

media-autobuild_suite.bat
1480c1: I'll try again to see if I can combine changes into logical commits

wiiaboo:
I've only looked at the changes, haven't tested anything, Looks fine so far.
Will probably be able to test only tomorrow after work.

media-autobuild_suite.bat
```diff
@@ -1354,142 +1337,156 @@ if not exist %instdir%\mintty.lnk (
...
+        echo.%instdir%\%msys2%\mingw32\ /mingw32
+        echo.%instdir%\%msys2%\mingw64\ /mingw64
+        if "%build32%"=="yes" echo.%instdir%\local32\ /local32
+        if "%build64%"=="yes" echo.%instdir%\local64\ /local64
+    )>"%instdir%\%msys2%\etc\fstab."
+)
+
 if not exist "%instdir%\%msys2%\home\%USERNAME%" mkdir "%instdir%\%msys2%\home\%USERNAME%"
 set "TERM="
+type nul >>"%instdir%\%msys2%\home\%USERNAME%\.minttyrc"
```
Helenerineium: Since you overwrite this file every time, the check below is redundant.

1480c1: .minttyrc is not actually overwritten, rather `type nul >> file` is like touch file as it will create if it doesn't exist, but nothing else

Helenerineium: https://ss64.com/nt/syntax-redirection.html
I see.

1480c1: I did not realize that batch supports the `<` redirect. That's actually good to know

media-autobuild_suite.bat
wiiaboo: If it closes on its own and doesn't require closing the .bat so that pacman.exe is updated, should be fine.

media-autobuild_suite.bat
1480c1: Although I squashed the commit that actually remove the code from below, I can't seems to figure out a good way to separate out the update.log part and pacman. Need to learn more about git

1480c1: I realized I just needed to rebase and edit and reset and recommit

PR #1412, opened 2019-10-01
Disable _FORTIFY_SOURCE

Helenerineium:
See https://github.com/msys2/MSYS2-packages/commit/a7d37497ba6bffeafe075ffdec8b0588b2493d62 and  https://github.com/msys2/MINGW-packages/issues/5803

> Prior to the change in mingw-w64 it had no effect so there is no loss by removing it.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1672,7 +1672,7 @@ goto :EOF
     echo.ACLOCAL_PATH="${LOCALDESTDIR}/share/aclocal:${MINGW_PREFIX}/share/aclocal:/usr/share/aclocal"
     echo.PKG_CONFIG="${MINGW_PREFIX}/bin/pkg-config --static"
     echo.PKG_CONFIG_PATH="${LOCALDESTDIR}/lib/pkgconfig:${MINGW_PREFIX}/lib/pkgconfig"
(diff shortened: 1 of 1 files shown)


media-autobuild_suite.bat
```diff
@@ -1672,7 +1672,7 @@ goto :EOF
     echo.ACLOCAL_PATH="${LOCALDESTDIR}/share/aclocal:${MINGW_PREFIX}/share/aclocal:/usr/share/aclocal"
     echo.PKG_CONFIG="${MINGW_PREFIX}/bin/pkg-config --static"
     echo.PKG_CONFIG_PATH="${LOCALDESTDIR}/lib/pkgconfig:${MINGW_PREFIX}/lib/pkgconfig"
-    echo.CPPFLAGS="-D_FORTIFY_SOURCE=2 -D__USE_MINGW_ANSI_STDIO=1"
+    echo.CPPFLAGS="-D_FORTIFY_SOURCE=0 -D__USE_MINGW_ANSI_STDIO=1"
```
```suggestion
    echo.CPPFLAGS="-D__USE_MINGW_ANSI_STDIO=1"
```
Wouldn't it just be better to remove the whole line section?

Helenerineium: No, flac (and possibly others) adds `-D_FORTIFY_SOURCE=2` if `_FORTIFY_SOURCE` is not defined.

So the build will fail with many undefined references (if updated to a newer mingw-w64 version that has support for `_FORTIFY_SOURCE` )

PR #1421, opened 2019-10-09
curl: add libidn2 and libpsl support

Helenerineium:


build/media-suite_compile.sh
```diff
@@ -330,6 +330,35 @@ if [[ $curl = y ]]; then
     enabled mbedtls && curl=mbedtls
     [[ $curl = y ]] && curl=schannel
 fi
+
+_check=(libunistring.{,l}a)
+if [[ $mediainfo = y || $bmx = y || $curl != n ]] && ! files_exist "${_check[@]}" &&
+    do_wget -h a82e5b333339a88ea4608e4635479a1cfb2e01aafb925e1290b65710d43f610b \
```
```
07:19:00 ┌ libunistring-0.9.10  ............................ [Downloaded]
07:19:00 ├ Hash mismatch, file may be broken: a82e5b333339a88ea4608e4635479a1cfb2e01aafb925e1290b65710d43f610b != eb8fb2c3e4b6e2d336608377050892b54c3c983b646c561836550863003c05d7 || db08bb384e81968957f997ec9808926e
```

build/media-suite_compile.sh
```diff
@@ -330,6 +330,35 @@ if [[ $curl = y ]]; then
     enabled mbedtls && curl=mbedtls
     [[ $curl = y ]] && curl=schannel
 fi
+
+_check=(libunistring.{,l}a)
+if [[ $mediainfo = y || $bmx = y || $curl != n ]] && ! files_exist "${_check[@]}" &&
```
wiiaboo: If the checks are all the same, just place it in a single if.

wiiaboo: Also, are all these three libraries not available statically from msys2?

wiiaboo: There's no dependency tracking. libunistring is needed by what? curl? libidn2 is required by what? libpsl is required for curl, I assume, but this is not being made explicit anywhere.

I see these three libs do exist in msys. So what's the issue with using them there? If it's the static libs being broken, can they not be fixed in msys? I'd rather not fill the suite with additional maintenance burdens unnecessarily.

wiiaboo: curl is being built specifically and optionally by the user because it's an "end" executable/lib with selectable features. There's no option to not include libidn/libpsl. I don't personally care for them, for example. Why am I wasting time compiling things I don't care for?
I'm not saying to add a .ini option to not include these, but they should at least be disableable through checking curl configuration options, so the user can explicitely add a --without-{libidn2,libpsl} to not waste time on these.

build/media-suite_compile.sh
```suggestion
    do_wget -h eb8fb2c3e4b6e2d336608377050892b54c3c983b646c561836550863003c05d7 \
```

Helenerineium:
> File URLs are valid for at least 30 days and up to a year

wiiaboo:
What?

wiiaboo:
From what I investigated a bit, libunistring should not need to be compiled. We just have to remove dllimport from unistring/woe32dll.h. libidn2 (IDN2_STATIC needs to be defined in its pkgconfig) and libpsl may need to be recompiled with dllimport removed from unistring as they still expect the symbols to contain the __imp_ prefix.

wiiaboo:
libunistring takes way too fucking long to compile, so that's definitely out of the question.

PR #1424, opened 2019-10-10
Empty array FFMPEG_OPTS before filling it

pat357:
When compiling the suite for both 32bit and 64 bit, the options from the 32 bit compile are mixed with those from the 64bit compile.
This is because when the suite is starting to compile for 64bit, the array containing the options is still filled with the options from the 32bit compile and the options for 64bit are just added on top.

This change will make the array empty before filling it with the options for 64bit to resolve this.

For MPV there is already a similar line MPV_OPTS=() in the code.

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -775,6 +775,7 @@ do_getFFmpegConfig() {
         fi
     fi
 
+    FFMPEG_OPTS=()
     for opt in "${FFMPEG_BASE_OPTS[@]}" "${FFMPEG_DEFAULT_OPTS[@]}"; do
         [[ -n $opt ]] && FFMPEG_OPTS+=("$opt")
     done


1480c1:
Please make sure to have a meaningful commit message

pat357:
What is the commit message ?
Is it the text "Empty array FFMPEG_OPTS before filling it" ?

If so, I agree with you, I should have chosen a more informative text.
Thanks for your feedback, I really appreciate this.
I've learned something new !

1480c1:
`Update media-suite_helper.sh`

1480c1:
Merged in https://github.com/m-ab-s/media-autobuild_suite/commit/2cfdafab9a10b259c902511425219928d81498b0

pat357:
Yes, now I see how stupid it looks !
I believe this is some kind of default message, so I have to figure out how to change it.
Can I still change it for this commit or is that no longer possible ?

1480c1:
I already changed it.

The usual way when using command-line git would be to

```bash
git fetch
git remote add upstream https://github.com/m-ab-s/media-autobuild_suite.git
git rebase upstream/master -i # Change the pick to reword
# git will open up the default editor and then edit the message to something appropriate and save
git push -f
```

But I'm not sure when doing from the website. Right now looking

1480c1:
Either that or

```bash
git commit --amend
git push -f
```

pat357:
When using the website, I can't give any git command's AFAIK.
I guess it must be a field somewhere that I forgot to edit.
I'll pay extra attention to it for my next pull request.

1480c1:
image
When editing from the website, make sure to change the first box as that is the commit message, make sure to have some kind of detail of what you are changing. I usually prefix the message with either the function name or file name and then the message, but really up to you

image

Looks like this in the end. (At the bottom)

pat357:
Thanks for the screenshots ! 
 I must have missed the field to edit the message, although it's clearly there..

Prefix the message with the function seems like a good idea.

PR #1427, opened 2019-10-11
Pat357check gnutls version online

pat357:
gnutls : check version online & trigger rebuild on update

Current code does not trigger a rebuild when the version is updated ; after updating the suite, one has to do a force rebuild to get the current version.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -330,11 +330,16 @@ if [[ $curl = y ]]; then
     enabled mbedtls && curl=mbedtls
     [[ $curl = y ]] && curl=schannel
 fi
-_check=(libgnutls.{,l}a gnutls.pc)
-if enabled_any gnutls librtmp || [[ $rtmpdump = y ]] || [[ $curl = gnutls ]] &&
-    ! files_exist "${_check[@]}" &&
-    do_wget -h b1f3ca67673b05b746a961acf2243eaae0ffe658b6a6494265c648e7c7812293 \
-    "https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/gnutls-3.6.10.tar.xz"; then
+
+if [[ $ffmpeg != no ]] && enabled_any gnutls librtmp || [[ $rtmpdump = y ]] || [[ $curl = gnutls ]]; then
+   [[ -z "$gnutls_ver" ]] &&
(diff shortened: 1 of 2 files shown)


pat357:
Dammed, now my pull request contains 2 patches, while I only wanted to have the one about updating gnutls. Please ignore the other patch as it was already added to the suite.
My apologies for this mistake.

build/media-suite_compile.sh
```diff
@@ -330,11 +330,16 @@ if [[ $curl = y ]]; then
...
+
+if [[ $ffmpeg != no ]] && enabled_any gnutls librtmp || [[ $rtmpdump = y ]] || [[ $curl = gnutls ]]; then
+   [[ -z "$gnutls_ver" ]] &&
+        gnutls_ver="$(clean_html_index "https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/")" &&
+        gnutls_ver="$(get_last_version "$gnutls_ver" "xz$" '3\.6\.\d+(\.\d+)?')"
+    gnutls_ver="${gnutls_ver:-3.6.10}"
+   _check=(libgnutls.{,l}a gnutls.pc) 
+    if do_pkgConfig "gnutls = $gnutls_ver"; then    
+        do_wget -h b1f3ca67673b05b746a961acf2243eaae0ffe658b6a6494265c648e7c7812293 \
+        "https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/gnutls-3.6.10.tar.xz"
```
1480c1: This may become a problem when a new gnutls version is released and doesn't match 3.6.10

pat357: Yes, indeed.
I'd consider to drop the hash code check and make it like this:
```
    if do_pkgConfig "gnutls = $gnutls_ver"; then    
        do_wget  "https://www.gnupg.org/ftp/gcrypt/gnutls/v3.6/gnutls-${gnutls_ver}.tar.xz"

Maybe replace the hash check by the .sig signatures available in the same dir https://www.gnupg.org/ftp/gcrypt/gnutls /v3.6/.

wiiaboo:
You could have also said somewhere that you're partiallly reverting 6334bdd06c8908e7307fab5bc877e6b2c045c17e as that's where the code is from.

wiiaboo:
Dynamic file integrity checking depends on #1221 being completed.

wiiaboo:
Meaning that, especially since this is a TLS library, I'd rather not change it back to dynamic version retrieval until integrity can also be dynamically checked.

PR #1433, opened 2019-10-14
curl: add libidn2 and libpsl support

Helenerineium:

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -369,26 +369,34 @@ fi
 { enabled mbedtls || [[ $curl = mbedtls ]]; } && do_pacman_install mbedtls
 
 if [[ $mediainfo = y || $bmx = y || $curl != n ]]; then
-    grep_or_sed dllimport "$MINGW_PREFIX"/include/unistring/woe32dll.h \
-    's|__declspec (dllimport)||g' "$MINGW_PREFIX"/include/unistring/woe32dll.h
+    do_pacman_install libunistring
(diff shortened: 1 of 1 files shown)


wiiaboo:
Looks better, I'll test it later today.

build/media-suite_compile.sh
```diff
@@ -406,9 +431,8 @@ if [[ $mediainfo = y || $bmx = y || $curl != n ]] &&
     hide_conflicting_libs
     CPPFLAGS+=" -DNGHTTP2_STATICLIB" \
         do_separate_confmakeinstall global "${extra_opts[@]}" \
-        --without-{libssh2,random,ca-bundle,ca-path,librtmp,libidn2} \
-        --with-brotli \
-        --enable-sspi --disable-{debug,manual}
+        --without-{random,ca-bundle,ca-path,librtmp} \
```
wiiaboo: You should keep libssh2 disabled. It's a troublesome lib to include in static builds.

wiiaboo:
- no such thing as do_curl
- used the new grep_and_sed instead of grep_or_sed which was doing nothing
- removed compilation and installation of lib test binaries if standalone is not enabled

Tested with openssl, mbedtls and schannel enabled (didn't know you could have multiple SSL backends, but I don't think it's that useful to add right now). No idea about gnutls, someone else can test it later.

wiiaboo:
Merged to e7d0532fd37e419ed434ff91615c84b0cad41689

PR #1438, opened 2019-10-15
do_wget: check hash before download

Helenerineium:
check hash of the file (if present) before attempting to download a copy, this can save some time in some circumstances.

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -368,12 +368,16 @@ do_wget() {
     $notmodified && [[ -f $archive ]] && curlcmds+=(-z "$archive" -R)
     [[ $hash ]] && tries=3
 
+    if [[ -f $archive ]] && [[ $hash ]] && check_hash "$archive" "$hash"; then
+        $quiet || do_print_status prefix "${bold}├${reset} " "${dirName:-$archive}" "$green" "File up-to-date"
+        tries=0
+    fi
+
     while [[ $tries -gt 0 ]]; do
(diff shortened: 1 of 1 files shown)


build/media-suite_helper.sh
```diff
@@ -368,12 +368,16 @@ do_wget() {
     $notmodified && [[ -f $archive ]] && curlcmds+=(-z "$archive" -R)
     [[ $hash ]] && tries=3
 
+    if [[ -f $archive ]] && [[ $hash ]] && check_hash "$archive" "$hash"; then
```
wiiaboo:
```suggestion
    if [[ -f $archive ]] && [[ -n $hash ]] && check_hash "$archive" "$hash"; then
```
right?

1480c1: Both, it seems that bash will treat brackets without a flag as -n

Better for readability with the n flag though

1480c1: Although it probably would help in cases where `$hash` contains a part valid for the brackets
```bash
$ [[ $asd ]]; echo $?
1
$ asd="-z 123"
$ [[ $asd ]]; echo $?
0
```

wiiaboo:
If it was a string, but a variable may be anything? Or is it still
evaluated as "-n stuff", not expanded to "-n" "stuff"?

1480c1:
```bash
$ asd="-f qwe"
$ [[ $asd ]]; echo $?
0
```
seems like it does not attempt to reevaluate

PR #1444, opened 2019-10-21
mpv: correct angle warning message

Helenerineium:


1480c1:
I'll leave it to @wiiaboo since I'm not knowledgeable enough to confirm if this is true

wiiaboo:
Yeah, since it's only the headers that get compiled, not the angle lib.

Issue #1441, opened 2019-10-18
Error running git update for SVT-AV1


ycwan:
``` bash
22:57:34   Running git update for SVT-AV1...
22:57:43 ┌ SVT-AV1 git  ................................. [Updates found]
22:57:43 ├ Running uninstall...
22:57:43 ├ Running cmake...
22:57:52 ├ Running build...
Likely error (tail of the failed operation logfile):
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../Bin/Release/libSvtAv1Enc.a(selfguided_avx2.c.obj):selfguided_avx2.c:(.text+0x1361): undefined reference to `__memset_chk'
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../Bin/Release/libSvtAv1Enc.a(selfguided_avx2.c.obj):selfguided_avx2.c:(.text+0x1379): undefined reference to `__memset_chk'
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../Bin/Release/libSvtAv1Enc.a(EbCodingLoop.c.obj):EbCodingLoop.c:(.text+0x2fd7): undefined reference to `__memcpy_chk'
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../Bin/Release/libSvtAv1Enc.a(EbCodingLoop.c.obj):EbCodingLoop.c:(.text+0x3477): undefined reference to `__memcpy_chk'
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ../Bin/Release/libSvtAv1Enc.a(EbCodingLoop.c.obj):EbCodingLoop.c:(.text+0x372d): undefined reference to `__memcpy_chk'
[...]

PR #1447, opened 2019-10-23
SVT-AV1: Apply 717 patch if pr is still open and mergeable

1480c1:
Temp fix for #1441 to apply the pr while it's open and mergeable since it seems it may take some time for upstream to merge

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1347,6 +1347,8 @@ if [[ $bits = "32bit" ]]; then
 elif { [[ $svtav1 = y ]] || enabled libsvtav1; } &&
     do_vcs "https://github.com/OpenVisualCloud/SVT-AV1.git"; then
     do_uninstall include/svt-av1 "${_check[@]}" include/svt-av1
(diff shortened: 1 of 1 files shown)


1480c1:
Testing if built jq works atm, if not, going to check jq

wiiaboo:
We could always install jq, whether or not we build it, to not get dependent on non-broken builds.
It'll be needed for future features too.

1480c1:
Seems to work fine for now

wiiaboo:
Works fine because you have jq built, but it'll break for people who don't.

1480c1:
Need to add pacman_install jq if not jq

wiiaboo:
Just add jq to mingwpackages in the .bat.

1480c1:
> Just add jq to mingwpackages in the .bat.

done

1480c1:
Should we stop building jq and use msys2 provided one?

wiiaboo:
No, I want a static one for myself.

wiiaboo:
The libs don't need to be installed, now that I see it. Only the .exe.

PR #1448, opened 2019-10-25
Helper: Add get_signature, check_signature, and do_jq 

1480c1:
Add functions helpful for getting signatures for gpg and checking if a file is correct using a sig or asc file.

I might be able to add a check if the file starts properly, but that is already handled by gpg erroring out with non-zero

Split off from https://github.com/m-ab-s/media-autobuild_suite/pull/1221

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -2013,6 +2013,14 @@ create_cmake_toolchain() {
         printf '%s\n' "${toolchain_file[@]}" > "$LOCALDESTDIR"/etc/toolchain.cmake
 }
 
+get_signature() {
+    # get_signature 96865171 0F3BE490
+    # adds keys to gpg keychain for verifying
+    for keyserver in keys.openpgp.org pool.sks-keyservers.net keyserver.ubuntu.com pgp.mit.edu; do
+        gpg --keyserver "$keyserver" --receive-keys "$@" && break
+    done > /dev/null 2>&1
+}
+
 grep_or_sed() {
     local grep_re="$1"
     local grep_file="$2"
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -2067,6 +2067,24 @@ check_signature() {
     esac
 }
 
+do_jq() {
+    local jq_file="$jq_file" output_raw_string=true
+    # Detect if in pipe, useful for curling github api
+    if [[ ! -t 0 ]]; then
+        jq_file=/dev/stdin
+    elif [[ -f $1 ]]; then
+        jq_file="$1" && shift
+    fi
+    for a in "$@"; do
+        grep -q -- ^- <<< "$a" && output_raw_string=false
+    done
+    if $output_raw_string; then
+        jq -r "$*" < "$jq_file"
+    else
+        jq "$@" < "$jq_file"
+    fi
+}
+
 grep_or_sed() {
     local grep_re="$1"
     local grep_file="$2"
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -2021,6 +2021,52 @@ get_signature() {
     done > /dev/null 2>&1
 }
 
+check_signature() {
+    # check_signature -k 96865171 gnutls-3.6.8.tar.xz.sig gnutls-3.6.8.tar.xz
+    # check_signature -k 1528635D8053A57F77D1E08630A59377A7763BE6 http://libsdl.org/release/SDL2-2.0.10.tar.gz.sig SDL2-2.0.10.tar.gz
+    # run in the same directory as the files. Works with .sig and some .asc
+    # file needs to start with -----BEGIN PGP SIGNATURE-----
+    local key=()
+    while true; do
+        case $1 in
+        -k) key+=("$2") && shift 2 ;; # keys to retrieve using get_signature
+        --)
+            shift
+            break
+            ;;
+        *) break ;;
+        esac
+    done
+    local sigFile=$1
+    shift
+
+    # Get name of sig file
+    local sigFileName=${sigFile##*/}
+    sigFileName=${sigFileName:-"$(/usr/bin/curl -sI "$sigFile" | grep -Eo 'filename=.*$' | sed 's/filename=//')"}
+    [[ -z $sigFileName ]] && echo "Sig file not set" && return 1
+
+    # Download sig file if url/cp file if local file
+    if ! do_wget -c -r -q "$sigFile" "$sigFileName" && [[ -f $sigFile ]]; then
+        sigFile="$(
+            cd_safe "$(dirname "$sigFile")"
+            printf '%s' "$(pwd -P)" '/' "$(basename -- "$sigFile")"
+        )"
+        [[ ${sigFile%/*} != "$PWD" ]] && cp -f "$sigFile" "$sigFileName" > /dev/null 2>&1
+    fi
+
+    # Retrive keys
+    [[ -n ${key[0]} ]] && get_signature "${key[@]}"
+
+    # Verify file is correct
+    # $? 1 Bad sig or file integrity compromised
+    # $? 2 no-pub-key or no file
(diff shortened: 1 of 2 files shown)


build/media-suite_helper.sh
```diff
@@ -399,7 +399,7 @@ do_wget() {
             ;;
         esac
 
-        if check_hash "$archive" "$hash"; then
+        $quiet || if check_hash "$archive" "$hash"; then
```
1480c1: Unsure if this line is needed or not

build/media-suite_helper.sh
```diff
@@ -2062,6 +2062,18 @@ check_signature() {
...
+do_jq() {
+    local jq_file="$jq_file"
+    # Detect if in pipe, useful for curling github api
+    if [[ ! -t 0 ]]; then
+        jq_file=/dev/stdin
+    elif [[ -f $1 ]]; then
+        jq_file="$1" && shift
+    fi
+
+    jq -r "$*" "$jq_file"
```
wiiaboo: Why `$*` instead of `$@`?

1480c1: jq does not support more than 1 json filter, anything after gets taken as a file parameter, so `curl ... | jq '.assets[]' '.'` is taken as wanting to open `.` as a json file

wiiaboo: Maybe just always redirect stdout to jq instead then, instead of using a file?

wiiaboo: We might want to use filtering/commands on jq, or certain options instead of always `-r`.

1480c1: > Maybe just always redirect stdout to jq instead then, instead of using a file?
That's the point of the if statements at the top

build/media-suite_helper.sh
wiiaboo: Yeah, this is not working properly for 404s. And even worse, it's returning 0.

build/media-suite_helper.sh
```diff
@@ -2013,6 +2013,15 @@ create_cmake_toolchain() {
         printf '%s\n' "${toolchain_file[@]}" > "$LOCALDESTDIR"/etc/toolchain.cmake
 }
 
+get_signature() {
+    # get_signature 96865171 0F3BE490
+    # adds keys to gpg keychain for verifying
+    {
+        gpg -q --receive-keys "$@" ||
+            gpg -q --keyserver pgp.mit.edu --receive-keys "$@"
```
wiiaboo: Any reason for this keyserver instead of the newer `hkps://keys.openpgp.org`?

1480c1: None, just because I wasn't sure what keyserver gpg doesn't use by default other than sks and I hope I can trust mit to keep their server up

build/media-suite_helper.sh
1480c1: I might just remove this until it's actually a problem

build/media-suite_helper.sh
wiiaboo: Yeah, but I'd rather try to use a newer keyserver which is not vulnerable to the key poisoning stuff.

build/media-suite_helper.sh
1480c1: Removed for now

build/media-suite_helper.sh
1480c1: Changed to first try `hkps://keys.openpgp.org`, then `hkps://hkps.pool.sks-keyservers.net`, then `hkps://pgp.mit.edu` in case the keys we need are not yet in `hkps://keys.openpgp.org` with the default server being `hkps://keys.openpgp.org`

build/media-suite_helper.sh
```diff
@@ -2022,6 +2022,46 @@ get_signature() {
...
+    if ! do_wget -c -r -q "$sigFile" "$sigFileName" && [[ -f $sigFile ]]; then
+        sigFile="$(
+            cd_safe "$(dirname "$sigFile")"
+            printf '%s' "$(pwd -P)" '/' "$(basename -- "$sigFile")"
+        )"
+        [[ ${sigFile%/*} != "$PWD" ]] && cp -f "$sigFile" "$sigFileName" > /dev/null 2>&1
+    fi
+
+    # Retrive keys
+    [[ -n ${key[0]} ]] && get_signature "${key[@]}"
```
wiiaboo: Good that this is optional.

build/media-suite_helper.sh
```diff
@@ -2022,6 +2022,46 @@ get_signature() {
...
+
+    # Verify file is correct
+    gpg -q --verify "$sigFileName" "$@" > /dev/null 2>&1
```
wiiaboo: Maybe add some user-friendly prints if in case gpg returns non-zero. Ideally, it should even error out the build, to be on the safe side.

build/media-suite_helper.sh
wiiaboo: I mean something like `< "$jq_file" jq -r -- "$@"`.

> jq does not support more than 1 json filter, anything after gets taken as a file parameter, so `curl ... | jq '.assets[]' '.'` is taken as wanting to open `.` as a json file

How is this an issue? It would also break `jq` itself. I don't think `"$@"` separates on spaces, so if you do `youtube-dl -J dQw4w9WgXcQ | jq -r '{title, album} | .title'` it wouldn't try to find a file called `album}`.
 
Speaking of which, `youtube-dl -J dQw4w9WgXcQ | jq -r '{title, album} | .title'` works with `jq` but not with `do_jq`.

1480c1: `< "$jq_file" jq -r "$@"` worked

1480c1: jq doesn't seem to have support for `--` parsing args stop

1480c1: > How is this an issue? It would also break `jq` itself. I don't think `"$@"` separates on spaces, so if you do `youtube-dl -J dQw4w9WgXcQ | jq -r '{title, album} | .title'` it wouldn't try to find a file called `album}`.

The `'{title, album} | .title'` would be taken as one due to the single quotes.

>@
Expands to the positional parameters, starting from one. When the expansion occurs within double-quotes, and where field splitting (see Field Splitting) is performed, each positional parameter shall expand as a separate field, with the provision that the expansion of the first parameter shall still be joined with the beginning part of the original word (assuming that the expanded parameter was embedded within a word), and the expansion of the last parameter shall still be joined with the last part of the original word. If there are no positional parameters, the expansion of '@' shall generate zero fields, even when '@' is double-quoted.

https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html

With `"$@"`, when running
```bash
youtube-dl -J dQw4w9WgXcQ | do_jq '{title, album} | .title' '|.'
```
`jq: error: Could not open file |.: Invalid argument`

With `$*`, when running 
```bash
youtube-dl -J dQw4w9WgXcQ | do_jq '{title, album} | .title'\
 '|.'
```
 `Rick Astley - Never Gonna Give You Up (Video)`

We can use newlines if it's `$*` without needing to keep the single quote open

wiiaboo: Yeah, but in what use case would you want `do_jq '{title, album} | .title' '|.'` to work?

wiiaboo: Ah, sorry, didn't read the last line.

wiiaboo: > ```shell
> youtube-dl -J dQw4w9WgXcQ | do_jq '{title, album} | .title\
>  |.'
> ```

Doesn't this work?

1480c1: Yes, it does

1480c1: > We might want to use filtering/commands on jq, or certain options instead of always `-r`.

Made it so that if there's an arg that starts with `-`, it will pass the options to jq as is `curl ... | do_jq -c .title` will run `jq -c .title` else it will run jq with `-r` on by default for easy bash parsing

wiiaboo:
Nice.

PR #1453, opened 2019-10-26
bat: add jo to compileArgs

Helenerineium:


1480c1:
What is jo for? Am I missing something?

wiiaboo:
It's a small bin to generate JSON from the shell.

PR #1454, opened 2019-10-30
do_pacman*: Fix actually skipping if not applicable

1480c1:
`pacman -Qqe` does not actually accept regex it seems, `pacman -Qqs` does accept regex though.

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1538,67 +1538,69 @@ do_unhide_all_sharedlibs() {
 }
 
 do_pacman_install() {
-    local pkg msyspackage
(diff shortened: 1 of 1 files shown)


1480c1:
I really wish we could have a way to test the PRs quickly. Sucks that it takes a really long time even with ccache (3 hours min) on github actions and github's new cache for actions is very limited in size to near uselessness

wiiaboo:
I can host a Win VM in a personal 24/7 server at home if it would help with automated testing.

1480c1:
I can too, but I'm not sure if there's an easy way to do something like that without maybe using gitlab runners.

Maybe I will look into that in my free time

PR #1456, opened 2019-10-31
gst: Add gstreamer

### LINKED ISSUE #1351: Request: gstreamer builds
I'm thinking of adding gstreamer as a package to be built so that way we can use the libs from the suite, but was wondering if there would be any interest/caveats to it?

Currently trying to see if I can replicate [gst-build](https://gitlab.freedesktop.org/gstreamer/gst-build) with the compile script, however [gstreamer](https://gitlab.freedesktop.org/gstreamer/gstreamer) fails to build due to PKG_CONFIG being exported as `pkg-config --static` since g-ir-scanner seems to ignore the PKG_CONFIG passed to meson (possibly revert f3347819e32ffc7ce7f6ae1e04bb686bc68aa397) 

Even with PKG_CONFIG set to pkg-config, since it doesn't pull the static flags, linking errors with undef refs to static glibc, ffi, iconv, and pcre symbols

1480c1:
Closes #1351

Currently builds everything static, I might need to add a system similar to ffmpeg and mpv of enabling and disabling since idk which plugins are non-free,gpl,lgpl etc yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ For information about the compiler environment see the wiki, there you also have
             - enabled by default, use --disable-sdl2 if unneeded
     - Zeranoe-emulating build (in addition to Light)
         - avisynth (needs avisynth dll installed)
-        - fontconfig (2.13.1)
+        - fontconfig (latest release)
         - only one of these TLS libs (including schannel) can be enabled at once:
             - openssl (mingw)
                 - preferred to gnutls and to libtls if all three are in options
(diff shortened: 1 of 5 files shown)


1480c1:
Follows the gst-build list with the exception of dev tools and stuff that relies on gobject-introspection since that only supports shared libs

wiiaboo:
That's a big boy feature. I'll try to get this and the others reviewed on the weekend or Monday.

1480c1:
Take your time, it may be possible that I might delay adding in gst-plugins-bad and gst-plugins-ugly since they might have licensing issues (One of the main reason why they are called bad and ugly)

wiiaboo:
On the license stuff, no idea, I don't know if people are going to use this, much less redistributing it. Try to at least figure out which gst-plugins-* are not GPL-compatible and put those behind a license=nonfree check.

Since this is static, we don't have the luxury of having the main app stay LGPL and allow the plugins to be nonfree.

1480c1:
Right now testing compilation again, if it works, I will merge

Issue #887, opened 2018-08-01
vpx git: make failed


rld243:
```
┌ vpx git ........................................... [Recently updated]
├ Running configure...
├ Running make...
Likely error:
{standard input}: Assembler messages:
{standard input}:19: Error: invalid register for .seh_savexmm
make[1]: *** [Makefile:157: vpx_dsp/x86/sad4d_avx512.c.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:17: .DEFAULT] Error 2
    [CC] vpx_dsp/x86/sad4d_avx512.c.o
{standard input}: Assembler messages:
{standard input}:19: Error: invalid register for .seh_savexmm
make[1]: *** [Makefile:157: vpx_dsp/x86/sad4d_avx512.c.o] Error 1
make: *** [Makefile:17: .DEFAULT] Error 2
make failed. Check D:/Tools/mabs/build/vpx-git/build-64bit/ab-suite.make.log
This is required for other packages, so this script will exit.
  Creating diagnostics file...

All relevant logs have been anonymously uploaded to https://0x0.st/sWfx.zip
Copy and paste [logs.zip](https://0x0.st/sWfx.zip) in the GitHub issue.
Make sure the suite is up-to-date before reporting an issue. It might've been fixed already.
Try running the build again at a later time.
```
[logs.zip](https://0x0.st/sWfx.zip)

BlackScreen:
Having the same error. Do we have to report it to the vpx git issues?

wiiaboo:
Yes.

rld243:
@Selur 
Yes, it fixed the problem. Thanks!

wiiaboo:
Ideally it should also be dealt with upstream.

PR #1461, opened 2019-11-05
vpx: re-enable avx512

rld243:
Disabling avx512 was a temporary workaround for https://github.com/m-ab-s/media-autobuild_suite/issues/887
It doesn't break compilation anymore, so It could be reverted I guess.


1480c1:
Have you completely tested it?

rld243:
FFmpeg with `--enable-libvpx` compiles with no errors. Created a few webms with it. All good.

Relevant commit:
https://chromium.googlesource.com/webm/libvpx/+/602d25ec4344f96031ce1043c15490e79438c1bc

PR #1465, opened 2019-11-07
bat: Add error if in MSVC environment

### LINKED ISSUE #1460: Fails to build fontconfig-git
OS: Windows 10 (1903) (x64)

Script fails on fontconfig-git

```
make[3]: Entering directory '/build/fontconfig-git/build-64bit/src'
lib -name:libfontconfig-1.dll -def:fontconfig.def -out:fontconfig.lib
make[3]: lib: Command not found
make[3]: *** [Makefile:986: fontconfig.lib] Error 127
```

The Msys environment doesn't appear to include a lib.
I tried using Microsoft's lib but that fails due to an invalid option.

[logs.zip](https://0x0.st/zg_U.zip)

1480c1:
Closes #1460

Should we only check for lib.exe or cl.exe? Is there some better way to detect running inside an MSVC environment?

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -68,6 +68,26 @@ do if %%f lss 4 (
     exit
 )
 
+(
+    where lib.exe || ^
+    where cl.exe || ^
+    if DEFINED VSINSTALLDIR cd .
+) >nul 2>&1 && goto MSVCINSTALLED
+goto MSVCNOTINSTALLED
+:MSVCINSTALLED
+echo ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


wiiaboo:
Dunno, I don't have a Windows installation with VS installed right now. But I'd say there's a few variables that are set by their .bats. At least something ending in CMN_TOOLS or COMNTOOLS or something.

wiiaboo:
https://renenyffenegger.ch/notes/Windows/development/Visual-Studio/environment-variables/index

wiiaboo:
But lib.exe and cl.exe should still not be in PATH, I think, unless some lib/package explicitely want them.

PR #1478, opened 2019-11-15
Codec2: svn->git

1480c1:
I don't think the symbols conflict was ever reported upstream

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -756,7 +756,7 @@ fi
 _check=(libcodec2.a codec2.pc codec2/codec2.h)
 if [[ $ffmpeg != "no" ]] && enabled libcodec2 && do_pkgConfig "codec2 = 0.8"; then
     [[ $standalone = y ]] && _check+=(bin-audio/c2{enc,dec,sim}.exe)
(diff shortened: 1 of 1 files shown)


wiiaboo:
It probably wasn't, but does this fix anything?

1480c1:
Just makes it so that the only svn repo is mplayer remaining

Would make it easier for me to possibly submit a patch or apply one to fix the conflicting mean symbol since I don't know how to use svn at all other than through `git svn`

wiiaboo:
But is it an official codec2 git repo or just a developer's sandbox?

1480c1:
I'm not entirely sure since there isn't any mention of it on the [sf page](https://sourceforge.net/projects/freetel/), however, it is mentioned on [freedv's page](https://freedv.org/#Source_Code), listed as `Codec 2 Source Code` and although I'm not entirely sure if the usernames match, a lot of the developers, including @drowe67 who is in the `Brought to you by` section, listed on the sf repository's history have been working on the github repository.

I don't think the sf repo page is abandoned as the mailing list seems active, although the latest commit was from 20 March 2019 compared to the git repo being 9 days ago and the git repository was also created 20 March 2019

So I'm not entirely sure if they moved all development to the github repository, but I think so

wiiaboo:
Seems like the github repo is where main development occurs, and sf is just
a stable release dump.
Sure, you can move it to git.

PR #1492, opened 2019-12-01
Resolve msys2-base extraction errors

ImportTaste:
Using dynamically defined variables as commands doesn't really work right without using call. Conditional execution is much more reliable.

You would think that starting the line with `7z >nul 2>&1 && ` on line 1359 would be unnecessary, but nope, conditional execution with piped commands is (usually) reliant on the first command in the chain not being piped. (There's more to it than that but it's not worth getting into.)

Also updated the instructions on what to do if the extraction fails, since it gives the impression you're supposed to put the msys2 folder in the build folder.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1355,10 +1355,8 @@ if not exist "%instdir%\%msys2%\msys2_shell.cmd" (
         echo.- Downloading and unpacking msys2 basic system
         echo.
         echo -------------------------------------------------------------------------------
-        7z >nul 2>&1 && set mabs7z=7z || set mabs7z=7za
-        7za >nul 2>&1 || powershell -NoProfile -NonInteractive -Command ^
-                ^(New-Object System.Net.WebClient^).DownloadFile^('https://github.com/chocolatey/chocolatey.org/raw/master/chocolatey/Website/7za.exe', '7za.exe'^)
-        %mabs7z% x msys2-base.tar.xz -so | %mabs7z% x -aoa -si -ttar -o..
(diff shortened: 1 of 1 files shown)


media-autobuild_suite.bat
```diff
@@ -1355,10 +1355,10 @@ if not exist "%instdir%\%msys2%\msys2_shell.cmd" (
         echo.- Downloading and unpacking msys2 basic system
         echo.
         echo -------------------------------------------------------------------------------
-        7z >nul 2>&1 && set mabs7z=7z || set mabs7z=7za
+        7z >nul 2>&1 || (
         7za >nul 2>&1 || powershell -NoProfile -NonInteractive -Command ^
-                ^(New-Object System.Net.WebClient^).DownloadFile^('https://github.com/chocolatey/chocolatey.org/raw/master/chocolatey/Website/7za.exe', '7za.exe'^)
-        %mabs7z% x msys2-base.tar.xz -so | %mabs7z% x -aoa -si -ttar -o..
+                ^(New-Object System.Net.WebClient^).DownloadFile^('https://github.com/chocolatey/chocolatey.org/raw/master/chocolatey/Website/7za.exe', '7za.exe'^))
```
```suggestion
                ^(New-Object System.Net.WebClient^).DownloadFile^('https://github.com/chocolatey/chocolatey.org/raw/master/chocolatey/Website/7za.exe', '7za.exe'^)
        )
```
For the bat file, since there are no auto-formatters and require formatting by hand and confirmation so it would be much better to close the parenthesis on the same column as the opening command for readability

1480c1: Why not just combine all of them instead of using parenthesis?

ImportTaste:
Done.

PR #1493, opened 2019-12-01
msys2-base.tar.xz cleanup once it's working

ImportTaste:
I noticed the file stuck around so I figured the best time to clean it up is after the second rebase. It definitely shouldn't be needed anymore after that.


ImportTaste:
My bad. Done.

Issue #1479, opened 2019-11-15
[mingw64] FFmpeg multiple definition of pathtools symbol


ycwan:
``` bash
23:56:39   Running git clone for ffmpeg...
23:58:33 ┌ ffmpeg git  ............................... [Recently updated]
23:58:33 ├ Until SVT-AV1 issues a fix, libsvtav1 must be disabled while libsvthevc is enabled....
23:58:33 ├ Changing options to comply to nonfree...
CUDA_PATH environment variable not set or directory does not exist.
patch unexpectedly ends in middle of line
23:58:43 ├ Compiling static FFmpeg...
23:58:47 ├ Running configure...
00:06:00 ├ Running pre make from vmaf_extra.sh...
00:06:02 ├ Running make...
Likely error (tail of the failed operation logfile):
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/mingw/msys64/mingw64/lib/libxml2.a(pathtools.o):(.text+0x6f0): multiple definition of `get_executable_path'; C:/mingw/msys64/mingw64/lib/libcrypto.a(pathtools.o):(.text+0x7e0): first defined here
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/mingw/msys64/mingw64/lib/libxml2.a(pathtools.o):(.text+0x7c0): multiple definition of `strip_n_prefix_folders'; C:/mingw/msys64/mingw64/lib/libcrypto.a(pathtools.o):(.text+0x8b0): first defined here
C:/mingw/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/mingw/msys64/mingw64/lib/libxml2.a(pathtools.o):(.text+0x820): multiple definition of `strip_n_suffix_folders'; C:/mingw/msys64/mingw64/lib/libcrypto.a(pathtools.o):(.text+0x910): first defined here
[...]

PR #1503, opened 2019-12-09
libxml2: Compile our own

1480c1:
Fixes https://github.com/m-ab-s/media-autobuild_suite/issues/1479 maybe

It seems most of the patches applied by msys2 is mainly just to fix warnings with the exception of the one that adds pathtools stuff.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -222,6 +222,19 @@ if [[ $dssim = y ]] &&
     do_checkIfExist
 fi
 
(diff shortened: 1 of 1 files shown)


wiiaboo:
Haven't tested it yet, but make sure there's no tools being built. We only need the lib itself.

1480c1:
Since they didn't seem to have an option to not build executables, I just added a step that runs their uninstall target with just deletes the executables after running the install

wiiaboo:
I usually patch the Makefile.am to not compile the binaries as well, to not waste time on stuff we don't need.

1480c1:
Done

PR #1508, opened 2019-12-23
ffmpeg+libvmaf: Replace default hard-coded model path

1480c1:
@wiiaboo, do you think there's a better way to do this?

Works after recompiling ffmpeg

Caused by https://github.com/Netflix/vmaf/issues/424

If there's nothing, I will merge this around tomorrow

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1875,6 +1875,10 @@ if [[ $ffmpeg != "no" ]]; then
                 }
         fi
 
+        # Replace default hard-coded model path with generated one based on where it's installed converted for ifstream to find
+        grep_and_sed '/usr/local/share/model/vmaf_v0.6.1.pkl' libavfilter/vf_libvmaf.c \
+            "s;/usr/local/share/model/vmaf_v0.6.1.pkl;$(cygpath -m "$LOCALDESTDIR/share/model/vmaf_v0.6.1.pkl");g"
+
         _patches="$(git rev-list origin/master.. --count)"
         [[ $_patches -gt 0 ]] &&
             do_addOption "--extra-version=g$(git rev-parse --short origin/master)+$_patches"


wiiaboo:
I'd prefer it in a relative path, if possible. If not, meh, it's a shit solution either way.

wiiaboo:
Also, this should be fixed in upstream, or at least suggested, so there's some other place in the Internet other than here in m-ab-s that this absolute path will work.

1480c1:
> I'd prefer it in a relative path, if possible. If not, meh, it's a shit solution either way.

Problem with a relative path would be that it would be relative to execution directory and would be the hardcoded default, which seems a bit weird that it's not a configure since it would be vendor specific be based on where vmaf is installed



> Also, this should be fixed in upstream, or at least suggested, so there's some other place in the Internet other than here in m-ab-s that this absolute path will work.
> 
> Absolute paths will also only work for the person that built it, and while they don't change the suite's root.

Yeah, I don't think there would be an easy way to solve this since this is compile time.

They would probably need to recompile to reset the default when they change the location. Probably the only thing that could help would be a configure flag for FFmpeg to change the default model path. This is mainly an issue on windows since we use `:` for valid absolute paths and indicating different drives, which ffmpeg's filter option consumes as separators.

wiiaboo:
You can escape colons with `\`, as normal, and then just use `/` instead of `\` for the path separators.
Frankly, I'd rather this stay broken as is, since at least it's broken everywhere the same way. Whereas if you hardcode a certain path, it might only work in a specific situation and it's not documented anywhere other than the suite's patching code.

1480c1:
Okay, I've already written to in one of their documentation issue to mention needing to escape colons and backslash for escaping the colon for Windows and I'll wait until then to ask to include that in ffmpeg's docs

1480c1:
I found a way to make it so it will load the vmaf module path relative to the ffmpeg executable, however, I don't think it's a perfect solution either way since that would depend on the user copying the local* folder instead of just copying ffmpeg.exe

PR #1513, opened 2019-12-30
Update README.md

aleksas:
`touch "$(get_first_subdir)/do_not_reconfigure"` raises `touch: cannot touch ... No such file or directory`. For ffmpeg_extra.sh "$(get_first_subdir)/do_not_reconfigure" resolves to `ffmpeg-git/do_not_reconfigure`, appaarently it should be `build/ffmpeg-git/do_not_reconfigure`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,10 +378,10 @@ _pre_configure(){
 
 _post_make(){
     # Don't run configure again.
-    touch "$(get_first_subdir)/do_not_reconfigure"
+    touch "$LOCALBUILDDIR/$(get_first_subdir)/do_not_reconfigure"
     # Don't clean the build folder on each successive run.
     # This is for if you want to keep the current build folder as is and just recompile only.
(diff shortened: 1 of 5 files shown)


1480c1:
> @1480c1 I thought you'd added the `/build` part as an option before. It could be either the default output of `get_first_subdir` and add an option to only print the dirname itself, or vice-versa if it's more sensible.
> There's a lot of places that I think are assuming get_first_subdir contains the absolute path, not just the readme.
> Besides the shit commit name, there's a lot of other places that need updating.

The thing I changed to print an abs path was `create_build_dir`. `get_first_subdir` never printed an abs path before. I could add that as an option since it would be helpful, something like either `get_first_subdir -f` to match `readlink` or `get_first_subdir -p`

1480c1:
Merged in https://github.com/m-ab-s/media-autobuild_suite/commit/f60db7971992a6851e0a1ca1496382b8615af3e9

PR #1515, opened 2019-12-30
bat: Rework INI logic

1480c1:
Does the logic here seem correct?

Also, would it make sense to adhere to `msys2ArchINI`? I feel that it might be better to have it as script runtime dependant rather than using a predefined value since it could be that someone migrates the suite from 32-bit to 64-bit or visa versa.


wiiaboo:
32-bit msys can run in 64-bit Windows, but 32-bit MSYS will always be broken as shit as none of the devs actually run it ever. The less chances we allow users to run 32-bit msys the better.

1480c1:
Should we eventually mark 32-bit as "depreciated"/"low-prio"?

wiiaboo:
It's already been that way for a while, so yeah.
Same as with 32-bit Windows.

PR #1530, opened 2020-01-16
aom: Add vmaf tune support

1480c1:
Tested and everything compiled fine

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -986,6 +986,18 @@ else
     pc_exists vpx || do_removeOption --enable-libvpx
 fi
 
+_check=(libvmaf.{a,pc} libvmaf/libvmaf.h)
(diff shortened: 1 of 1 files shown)


wiiaboo:
Isn't this only used by aomenc? Or also by libaom?

1480c1:
Only aomenc I believe

Selur:
aomenc offers: `--tune=vmaf` as option, but when that option is used one gets:
```
Failed to control codec: Invalid parameter
    This error may be related to the wrong configuration options: try to set -DCONFIG_TUNE_VMAF=1 at the time CMake is run.
```
-> would be nice it tune vmaf would be enabled in cmake

moisespr123:
The commit was reverted.

Selur:
any news on this?

PR #1534, opened 2020-01-17
rav1e: Use jobs argument for cinstall

1480c1:
@wiiaboo, is there a particular reason why we don't use cinstall directly to LOCALBUILDDIR with `--bindir`?


wiiaboo:
It installs shared crap which we don't need.

PR #1537, opened 2020-01-22
do_vcs: Some more cleanup and improvements(?)

1480c1:
Haven't fully tested if there's any speed improvements, will try too

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -135,20 +135,19 @@ vcs_get_default_ref() {
 }
 
 # vcs_get_current_type /build/myrepo
-vcs_get_current_type() {
-    local basedir
-    basedir=${1:-$(get_first_subdir -f)}
-    if [[ -d $basedir/.git ]] && git -C "$basedir" rev-parse --is-inside-work-tree > /dev/null 2>&1; then
-        echo "git"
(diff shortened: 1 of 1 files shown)


build/media-suite_helper.sh
```diff
@@ -135,20 +135,19 @@ vcs_get_default_ref() {
...
-    elif [[ -d $basedir/.hg ]] && hg -R "$basedir" id --id > /dev/null 2>&1; then
-        echo "hg"
-    elif [[ -d $basedir/.svn ]] && svn info --depth empty "$basedir@HEAD" > /dev/null 2>&1; then
-        echo "svn"
-    else
-        echo "unknown"
-        return 1
-    fi
-}
+vcs_get_current_type() (
```
wiiaboo: Does this refactored version work if more than one of the directories exist at the same time?
The previous version also tested some vcs-specific commands for validity before jumping to the next possibility.

1480c1: Yes since it will loop over each dot file and only return 0 if the dot file matched one and the vcs specific command ran successfully

wiiaboo: So the case will loop through every `^.*` file it finds in the current directoryuntil it finally ends?
Also, it'll also match for a `.git`. But I guess since the commands should fail in those cases, that still works.

wiiaboo: Wait, nevermind, the case is inside the for loop, case just runs once per file.

PR #1557, opened 2020-02-14
Zab-suite.sh: Source helper.sh

1480c1:
Helps me not have to write
```bash
source /build/media-suite_helper.sh
```
every single time I ask somebody to run a command in mintty

Is there a particular reason this was not done before?

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1601,6 +1601,9 @@ findstr /C:"profile2.local" %instdir%\%msys2%\etc\profile.d\Zab-suite.sh >nul 2>
     echo.elif [[ -z "$MSYSTEM" ^|^| "$MSYSTEM" = MINGW32 ]]; then
     echo.   source /local32/etc/profile2.local
     echo.fi
+    echo.if [[ -f /build/media-suite_helper.sh ]]; then
+    echo.   source /build/media-suite_helper.sh
+    echo.fi
 )>%instdir%\%msys2%\etc\profile.d\Zab-suite.sh
 
 rem compileLocals


wiiaboo:
Because I had it on my .bashrc for myself. There's no way to unsource an unneeded script, so I'm not sure if wouldn't be better to add it to the mintty shortcut somehow instead.

wiiaboo:
Or source it depending on a magic environment variable being set.

1480c1:
I see, there doesn't seem to be a good way to make it so that the shortcut that we generate will auto source helper.sh and I don't think there's a good way to make it so that it will take a magic env variable unless we use `/usr/bin/env` and pass it there

1480c1:
Also on an annoying side note, shfmt in vs code all of the sudden started working in the compile script, so now I have to click cancel whenever it tries to save, else I have all of the code tabbing itself

wiiaboo:
You can't use a specific configuration with shfmt like with eslint?

wiiaboo:
We could also use a magic env var to not source the rest of helper.sh. Then you could "disable" it and basically source almost none of it?

1480c1:
> You can't use a specific configuration with shfmt like with eslint?

Although it does provide a way to add flags to shfmt, shfmt does not have a flag to exclude certain files. VS code does not provide a way to disable formatting for a specific file https://github.com/Microsoft/vscode-python/issues/2776#issuecomment-427932938
> You can now use the command Save Without Formatting in the new release of VS Code.
> Considering a solution is available in VS Code, we're reluctant to provide a custom solution in the Python Extension at this stage. Hence I'm closing this issue.

https://stackoverflow.com/questions/44831313/how-to-exclude-files-from-format-on-save-in-vscode

Although, after typing all of this, I found https://github.com/mvdan/sh/issues/288 and it seems they made it so you can exclude using a `.editorconfig` with
```toml
[build/media-suite_compile.sh]
ignore = true
```
so it seems I was looking in the wrong places for shfmt options



> We could also use a magic env var to not source the rest of helper.sh. Then you could "disable" it with the env var and have an if at the start and basically source almost none of it? I'm not sure if you can do that with bash.

I'm not sure if that would be a good idea either since it would be hard to detect an intent to source it by a user and differentiate that from sourcing from a profile and sourcing from the script context easily

1480c1:
Nvm on the editorconfig, it doesn't seem to work.

PR #1568, opened 2020-03-03
bat: Add clang option

### LINKED ISSUE #802: Clang support
Just wondering, how much effort could it take to switch from mingw/gcc to Clang compiler?
Could it be done by just replacing compiler executable somewhere? (assuming that Clang attempts to copy the command-line options from gcc)
The reason I want it is to have debug symbols in PDB format to use with Visual Studio.

1480c1:
Closes #802


1480c1:
merged in https://github.com/m-ab-s/media-autobuild_suite/commit/c304889e3c3d8c5ec6741d44ac7992a8a80018f1

Issue #1512, opened 2019-12-30
Request: Add ability to reproduce build including git revisions and other 3-rd party tool versions


aleksas:
To comply with LGPL build being reproducible is important, so it would be helpful if there would be option to not just build or update according to media-autobuild_suite.ini but also reproduce the build with frozen revisions and tool versions.

wiiaboo:
It's not the suite that needs to comply with LGPL, it's the people that redistribute binaries compiled using it.

But I agree it would be somewhat useful to centralize the VCS commit hashes/versions somewhere, maybe more useful than the changelog file.
And patches need to be considered/gathered easily. If you apply `git am` patches it'll change the commit that goes into the compiled binary. If you apply `patch` patches it'll be a "dirty" build. Stuff like that.

There's no point in documenting msys dependencies as we have no control over them or what patches they use, so only document what we can, from dependencies built by the suite.

1480c1:
I have been considering adding a functionality such as maybe something `repo_hashes.txt` to add in `<reponame>=<hash>` to checkout whatever is at `<hash>` instead of whatever the default is.

An alternative would be a json file that could contain options such as changing the url to pull, commit to checkout, and additional patches, similar to pacman's pkgbuild, but that would require a bit of time.

1480c1:
Perhaps I can link the packages labeled with `mingw` on the readme with a link to their folder in msys2/MINGW-packages

wiiaboo:
There's no guarantee that MINGW-packages matches absolutely what they use to build. They might not've built the most recent version of the package, or it might not exist in git anymore but still be available as a package.

PR #1588, opened 2020-03-24
Specific FFmpeg version is supported.

nineKnight:


media-autobuild_suite.bat
```diff
@@ -688,6 +688,26 @@ if %buildffmpegUp%==3 set "ffmpegUpdate=onlyFFmpeg"
...
 if %deleteINI%==1 echo.ffmpegUpdate=^%buildffmpegUp%>>%ini%
 
+:ffmpegVer
+if %ffmpegVersionINI%==0 (
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+    echo.
+    echo. Always build latest FFmpeg?
+    echo. 1 = Yes
+    echo. x.y.z = Specific FFmpeg version (the git tag, e.g. 4.2.2 for n4.2.2^)
```
wiiaboo: No, this seems like a terrible ideia.
There's no verification at all of what the user writes.

build/media-suite_config.sh
```diff
@@ -0,0 +1,30 @@
...
+  case $1 in
+    --ffmpegVersion=* ) ffmpegVersion="${1#*=}"; shift ;;
+    -- ) shift; break ;;
+    -* ) echo "Error, unknown option: '$1'."; exit 1 ;;
+    * ) break ;;
+  esac
+done
+
+[ "$ffmpegVersion" == '' ] && ffmpegVersion='latest'
+sed -ri '/readme.txt/{s/(ffmpeg-).*(-win)/\1'$ffmpegVersion'\2/g}' "$LOCALBUILDDIR"/media-suite_helper.sh
```
wiiaboo: No.
Why would you need to change suite files for this? This would break git-synced suite installations.

build/media-suite_compile.sh
```diff
@@ -1916,7 +1927,7 @@ if [[ $ffmpeg != "no" ]]; then
             "$LOCALDESTDIR"/bin-video/{sw{scale,resample},postproc}.lib
             )
         _check=()
-        sedflags="prefix|bindir|extra-version|pkg-config-flags"
+        sedflags="prefix|bindir|extra-\w*|pkg-config-flags|disable-debug|shlibdir"
```
wiiaboo: Why are you removing disable-debug? This is imperative to leave in the compiled binary for proper issues to reported to FFmpeg.
These modifications are unrelated to this PR and should proposed in a separate PR or issue.

build/media-suite_compile.sh
```diff
@@ -1438,6 +1439,11 @@ if [[ $x264 != no ]]; then
         if [[ $standalone = y && $x264 =~ (full|fullv) ]]; then
             _check=("$LOCALDESTDIR"/opt/lightffmpeg/lib/pkgconfig/libav{codec,format}.pc)
             do_vcs "https://git.ffmpeg.org/ffmpeg.git"
+            if [[ $ffmpegVersion == latest ]]; then
+                echo > VERSION
```
wiiaboo: No, this is also a terrible idea. Why are you forcing FFmpeg to use what's inside this file, which might not even match in anyway the current checked out branch/tag?

wiiaboo:
A proposed solution to this was already thought about in https://github.com/m-ab-s/media-autobuild_suite/issues/1512#issuecomment-569697115.

Overall, this PR seems like a terrible implementation of that idea.

media-autobuild_suite.bat
nineKnight: No guarantee is needed. `do_vcs` will fail and log will be written.

build/media-suite_config.sh
nineKnight: Is there a script for the suite to update?

build/media-suite_compile.sh
nineKnight: You can leave it as you wish.

build/media-suite_compile.sh
nineKnight: Do you really verify how it works?

nineKnight:
@wiiaboo Your version is desired. Appreciate your work.

media-autobuild_suite.bat
wiiaboo: Yeah, the user does stupid things and we have to deal with it. That's not how it works.

build/media-suite_config.sh
wiiaboo: Yes, and there's the fact that it's git.

build/media-suite_compile.sh
wiiaboo: Not 100%, but VERSION is used for release branches by FFmpeg, so anyway it would be wrong to write to it.

wiiaboo:
This is a low-priority feature that will be implemented when someone else has time and interest for it. That's what you pay for with using open source.

build/media-suite_compile.sh
nineKnight: If `echo > VERSION`, FFmpeg won't use the `VERSION` file. So if not do it like this, how do you specify the version shown in `ffmpeg -version`, or do you always build the latest FFmpeg?

wiiaboo: --extra-version, like I said somewhere else.

nineKnight:
Yes, you are right.

在 2020年3月27日星期五，Ricardo Constantino <notifications@github.com> 写道：

Barough:
Think it would be great if someone could make it possible to compile a specific ffmpeg version through the suite

build/media-suite_compile.sh
nineKnight: Yes, you are right.

PR #1595, opened 2020-04-01
Update Python and Vapoursynth

Patman86:
Update Python to 3.8.2 and Vapoursynth to R49


1480c1:
Have you tested if this works?

1480c1:
Going to assume it works since the remote files exist

Patman86:
Yes, I've tested it and it works 😉

PR #1604, opened 2020-04-05
.bat: temporarily disable avisynth

wiiaboo:


wiiaboo:
Just the temporary stopgap, solution will be to install https://github.com/AviSynth/AviSynthPlus headers.

PR #1617, opened 2020-04-15
libaom: Move libaom to light build from zeranoe

1480c1:
Since we already have libdav1d for decoding av1, it would be good to also have a library to encode av1.

@wiiaboo any comments?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ For information about the compiler environment see the wiki, there you also have
         - cuda-llvm (built-in)
         - cuvid (built-in)
         - ffnvcodec (git)
+        - libaom (git)
         - libdav1d (git)
         - libmp3lame (mingw)
         - libopus (mingw)
@@ -49,7 +50,6 @@ For information about the compiler environment see the wiki, there you also have
             - mbedtls (mingw)
                 - preferred to gnutls if GPLv3 license is chosen
             - gnutls (latest release)
-        - libaom (git)
         - libass (git)
             - by default with DirectWrite backend
             - if --enable-fontconfig, fontconfig backend included
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -103,14 +103,14 @@ iconv lzma nvenc schannel zlib sdl2 ffnvcodec nvdec cuda-llvm
 
 :: common external libs
 set ffmpeg_options_basic=gmp libmp3lame libopus libvorbis libvpx libx264 libx265 ^
-libdav1d --disable-debug
+libdav1d libaom --disable-debug
 
 :: options used in zeranoe builds and not present above
 set ffmpeg_options_zeranoe=fontconfig gnutls libass libbluray libfreetype ^
 libmfx libmysofa libopencore-amrnb libopencore-amrwb libopenjpeg libsnappy ^
 libsoxr libspeex libtheora libtwolame libvidstab libvo-amrwbenc libwavpack ^
 libwebp libxml2 libzimg libshine gpl openssl libtls avisynth mbedtls libxvid ^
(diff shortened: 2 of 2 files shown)


wiiaboo:
dav1d is there because it's a light lib for decoding av1 (for which there's no native decoder in ffmpeg yet).

~~You can add aom since you still need to actually agree to compile it in the .ini. vpx/x264/x265 are also there.~~
libaom is also compiled everytime if you enable it in ffmpeg, unlike the other encoders.
I would vote for having the same behavior as the other encoders.

It's still a painfully slow encoder for most CPUs, afaik? Are people actually already encoding using it?

1480c1:
> dav1d is there because it's a light lib for decoding av1 (for which there's no native decoder in ffmpeg yet).
> 
> ~You can add aom since you still need to actually agree to compile it in the .ini. vpx/x264/x265 are also there.~
> libaom is also compiled everytime if you enable it in ffmpeg, unlike the other encoders.
> I would vote for having the same behavior as the other encoders.

Sure, I can go through the rest of the projects in the light build etc and make sure they have similar `enabled x` conditionals

> 
> It's still a painfully slow encoder for most CPUs, afaik? Are people actually already encoding using it?

afaik, for encoding to av1, there is only really libaom-av1 and librav1e since libsvtav1enc is currently a bit too unstable to really use, especially since it's not mainlined and the quality can be considerably worse compared to the first two, so I would not recommend to add it to the  light build And out of the first I only libaom really satisfies the stability since librav1e's C-API can easily break based on multiple projects, not just rav1e.

Plus it seems there are some ~weebs~ people already encoding to av1 https://nyaa.si/?f=0&c=0_0&q=av1

Monteco:
There is an active community for AV1 encoding, and we really enjoy MABS tool. It would be great for libaom-av1 to be included and I also must point out that it is as fast as HEVC at very slow speeds (However due to poor multithreading you have to launch multiple encodes at once).

master-of-zen:
> It's still a painfully slow encoder for most CPUs, afaik? Are people actually already encoding using it?

@wiiaboo Yes.
I'm developer of [Av1an](https://github.com/master-of-zen/Av1an), and with my project people with 3970X were getting ~8 fps for 1080p 10b content with cpu-used 3 and ~25fps on cpu-used 5. Which means that it's possible to encode whole movie with great settings overnight.
Threadripper showcase with encoded examples [link](https://github.com/master-of-zen/Av1an/releases/tag/1.6).

wiiaboo:
> There is an active community for AV1 encoding, and we really enjoy MABS tool. It would be great for libaom-av1 to be included and I also must point out that it is as fast as HEVC at very slow speeds (However due to poor multithreading you have to launch multiple encodes at once).

It is already present in the suite, just not as part of the light preset.

> > It's still a painfully slow encoder for most CPUs, afaik? Are people actually already encoding using it?
> 
> @wiiaboo Yes.
> I'm developer of [Av1an](https://github.com/master-of-zen/Av1an), and with my project people with 3970X were getting ~8 fps for 1080p 10b content with cpu-used 3 and ~25fps on cpu-used 5. Which means that it's possible to encode whole movie with great settings overnight.
> Threadripper showcase with encoded examples [link](https://github.com/master-of-zen/Av1an/releases/tag/1.6).

"Overnight" reminds me of the DivX/XviD times when people used GKnot.
For quality-centric encodes I expect AV1 to already match or top HEVC, sure, but most people that would use the suite without most of its functionality only use mpv/ffmpeg for quick encodes, probably. While 25fps seems close to realtime, a 3970X is not really a "common" CPU.

But sure, add libaom to the light preset.

1480c1:
I will work on looking at the other encoders to make them match

PR #1668, opened 2020-05-22
bat: Drop support for 32-bit hosts

1480c1:
Since https://github.com/msys2/msys2-installer/releases doesn't have any i686 sfx.exe releases, I think this should be fine to merge. This just removes the ability to run the compile suite on 32-bit host machine, this does not affect mingw-w64-i686-* packages, so for now 32-bit can be compiled, but not compiled on.


wiiaboo:
Could you add some type of warning/error exit to prevent people from running the suite in 32-bit hosts?

wiiaboo:
Also, update the Readme to remove references to the option being available and a strict requirement for 64-bit hosts.

1480c1:
Will do really quickly

1480c1:
Done

PR #1669, opened 2020-05-23
Bat: Add low prio option

1480c1:
Might help with people who want to compile in the background, we could add more priority options too

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -1694,7 +1694,7 @@ if %noMintty%==y (
     call :runBash compile.log /build/media-suite_compile.sh %compileArgs%
 ) else (
     if exist %build%\compile.log del %build%\compile.log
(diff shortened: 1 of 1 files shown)


wiiaboo:
Does this actually do anything? (honestly curious)

1480c1:
It sets mintty and bash and spawned process to low priority, so at least i can play games while a compile is running in the background without it freezing or lagging my game

wiiaboo:
Couldn't you just use lower cpuCount?

wiiaboo:
I'd even agree to always run the suite on low priority as a base, since if you're not doing anything else, it'll probably compile just as fast.

1480c1:
Yeah, the main reason to have this set to on would be so during lull times the compile will run at full speed, but can be backgrounded if there's something more pressing.

Should i set this to default?

wiiaboo:
Sure.

PR #1672, opened 2020-05-25
Mention libsvtvp9 in README.md

ImportTaste:
I noticed #1619 forgot to mention this in the documentation, so here it is.


ImportTaste:
Oh, that might be intentional since 0001-Add-ability-for-ffmpeg-to-run-svt-vp9.patch doesn't seem to apply currently.

PR #1673, opened 2020-05-25
Allow --enable-librav1e

ImportTaste:
As far as I can tell, this is no longer necessary. Built and tested it, it works just fine.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1903,11 +1903,6 @@ if [[ $ffmpeg != no ]]; then
         enabled vapoursynth &&
             do_patch "https://0x0.st/zp4W.txt vapoursynth_alt.patch" am
 
-        # librav1e
-        if enabled librav1e; then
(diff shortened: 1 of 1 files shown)


1480c1:
Did you compile ffmpeg shared? This check only removes librav1e from ffmpeg when building ffmpeg shared

ImportTaste:
Yes, I did.

1480c1:
Okay, i will test as well to make sure in the morning

1480c1:
Can confirm that a shared ffmpeg build will work fine with librav1e enabled and I can encode using `-c:v librav1e`

wiiaboo:
Sure, just change the commit to reflect that it's only now being allowed for shared.

ImportTaste:
There, I think that's what you wanted, if not, let me know what you'd rather the commit message be.

wiiaboo:
Just `ffmpeg: allow librav1e for shared builds` would be fine, if you still want to change it yourself.

ImportTaste:
Done

PR #1705, opened 2020-06-09
Add SVT-VP9 temporary workaround

ImportTaste:
https://github.com/m-ab-s/media-autobuild_suite/pull/1619#issuecomment-616206347 taught me that ffmpeg cannot be compiled with SVT-VP9 while the SVT-AV1 and XVid options are enabled (and I did test to make sure XVid did indeed conflict). There doesn't seem to be anything to reference or handle this conflict other than these few comments, so I made a quick codeblock to do so.

I also revised the SVT-AV1 codeblock in the process. Much cleaner code seems a bit more important than trying to keep the log message down to a single line.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1915,6 +1915,16 @@ if [[ $ffmpeg != no ]]; then
         _deps=(lib{aom,tesseract,vmaf,x265,vpx}.a)
     if do_vcs "https://git.ffmpeg.org/ffmpeg.git"; then
 
+        # ((TEMPORARY SVT-VP9 MEASURES))
+        # Reasons for this codeblock = https://github.com/m-ab-s/media-autobuild_suite/pull/1619#issuecomment-616206347
+        if enabled libsvtvp9; then
+            enabled libxvid && do_removeOption --enable-libxvid &&
+                do_print_progress "Until an upstream fix is issued, compiling with libsvtvp9 must disable libxvid."
+            enabled libsvtav1 && do_removeOption --enable-libsvtav1 &&
+                do_print_progress "Until an upstream fix is issued, compiling with libsvtvp9 must disable libsvtav1."
+        fi
+        # (/(TEMPORARY SVT-VP9 MEASURES))
+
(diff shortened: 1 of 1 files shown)


wiiaboo:
Lol. I'd sooner prefer disabling svt-vp9. That's a lot of broken shit because of one lib.

ImportTaste:
They're different projects, SVT-VP9 is incompatible with two libraries and SVT-AV1 is incompatible with two libraries.

wiiaboo:
Still, why are SVT libs incapable of not breaking other libs?

1480c1:
For svt-av1, it's because they copied a lot of aom's code verbatim, for the rest, idk

ImportTaste:
I'm guessing SVT-VP9 copied some AV1 code and it thankfully didn't conflict with libaom and that other one. The xvid conflict is probably the most confusing.

wiiaboo:
The opencore-amrwb is the most confusing, it's an audio codec.

ImportTaste:
Huh, you're right. And apparently it's only used for decoding. The other opencore one is for encoding.

1480c1:
iirc, the conflict with opencore-amrwb was some kind of short, and very common name.

https://github.com/OpenVisualCloud/SVT-AV1/issues/567

moise noted it https://github.com/OpenVisualCloud/SVT-AV1/issues/567
`pred`

wiiaboo:
Nice, a function probably used by every codec, except most of them prefix it with something so it doesn't conflict. I'm not expecting a dead lib to change it, but SVT is probably not dead and they could easily change it, no?

1480c1:
I have a feeling the `pred` one is already solved in SVT-AV1, but just not in SVT-VP9, and SVT-HEVC is safe because they used camelcasing or something I think

wiiaboo:
`pREd`?

1480c1:
looks like the current quality of the code atm, lol

greping the output of nming the hevc lib for `-i ' pred'` doesn't bring any symbol with just `pred` up, I thought it was something like `Pred`

PR #1708, opened 2020-06-12
Make the suite continue if exhale fails to build

moisespr123:
Exhale isn't really needed for other components, so we can make it not required by the suite and let it continue shall it fail to build.

In the future, if the author makes an ffmpeg extension, this can be reverted.


1480c1:
commited in https://github.com/m-ab-s/media-autobuild_suite/commit/5939237f21177e437ffb91d4a60fdc48dd029f19

Issue #1725, opened 2020-06-26
64 bit, Error running make when updating ffmpeg git–[/build/ffmpeg-git/ffbuild/common.mak:59: libavformat/libsrt.o] Error 1


ycwan:
``` bash
23:30:55 ┌ ffmpeg git  .................................. [Updates found]
23:30:55 ├ Until an upstream fix is issued, libaom must be disabled when compiling with libsvtav1....
23:30:55 ├ Until an upstream fix is issued, libopencore-amrwb must be disabled when compiling with libsvtav1....
23:30:55 ├ Changing options to comply to nonfree...
CUDA_PATH environment variable not set or directory does not exist.
23:31:09 ├ Compiling static FFmpeg...
23:31:11 ├ Running uninstall...
23:31:11 ├ Running configure...
23:41:10 ├ Running make...
Likely error (tail of the failed operation logfile):
      |                                                  SRTO_SENDER
src/libavformat/libsrt.c: In function 'libsrt_setup':
src/libavformat/libsrt.c:409:5: warning: 'srt_socket' is deprecated [-Wdeprecated-declarations]
  409 |     fd = srt_socket(cur_ai->ai_family, cur_ai->ai_socktype, 0);
      |     ^~
In file included from src/libavformat/libsrt.c:24:
C:/MinGW64/local64/include/srt/srt.h:735:41: note: declared here
  735 | SRT_ATR_DEPRECATED_PX SRT_API SRTSOCKET srt_socket(int, int, int) SRT_ATR_DEPRECATED;
      |                                         ^~~~~~~~~~
make: *** [/build/ffmpeg-git/ffbuild/common.mak:59: libavformat/libsrt.o] Error 1
make failed. Check C:/MinGW64/build/ffmpeg-git/build-static-64bit/ab-suite.make.log
This is required for other packages, so this script will exit.
23:46:34   Creating diagnostics file...

All relevant logs have been anonymously uploaded to https://0x0.st/iJPx.zip
Copy and paste [logs.zip](https://0x0.st/iJPx.zip) in the GitHub issue.
Make sure the suite is up-to-date before reporting an issue. It might've been fixed already.
Try running the build again at a later time.

```

[logs.zip](https://0x0.st/iJPx.zip)
[ab-suite.make.log](https://github.com/m-ab-s/media-autobuild_suite/files/4838008/ab-suite.make.log)

1480c1:
https://github.com/Haivision/srt/issues/1382

[...]

PR #1730, opened 2020-07-02
SVT_VP9: No longer disables xvid/svtav1 for ffmpeg

ImportTaste:
Did some test compiles today using the patch described here until it's corrected upstream: https://github.com/m-ab-s/media-autobuild_suite/issues/1725#issuecomment-652664267

I can confirm that SVT_VP9 no longer needs to disable libxvid or libsvtav1. I compiled both shared and static builds successfully.

_(Unfortunately, libsvtav1 still needs to disable libaom and libopencore-amrwb)_


ImportTaste:
I might be an idiot. I will re-open this PR if it turns out I am, in fact, not.

PR #1758, opened 2020-07-19
do_vcs: drop mercurial and svn

1480c1:
Wanted to get someone else to also look at this code before I merge it just to make sure.

ideally, `do_vcs` will
```bash
if ! proper repo; then
    clone repo
fi
fetch updates, but not merge
get the merge-base of the current local repo and upstream
if not checked recently; then
    reset to upstream
fi

check if merge base != current hash and return new updates
rest of the logic stuff
```

for mplayer, I just put a static version of `do_vcs` with the unnecessary stuff removed, so as long as svn doesn't change or anything, it shouldn't need to be changed. I am considering just maybe moving the actual code into a subshell inside the if, but that also seems like it would cause a lot of ugliness.

maybe another time I might actually look a bit more at cleaning up the mplayer portion

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2049,7 +2049,7 @@ if [[ $ffmpeg != no ]]; then
 fi
 
 # static do_vcs just for svn
-do_svn() {
+check_mplayer_updates() {
     cd_safe "$LOCALBUILDDIR"
     if [[ ! -d mplayer-svn/.svn ]]; then
         rm -rf mplayer-svn
@@ -2111,7 +2111,7 @@ do_svn() {
 }
 
 _check=(bin-video/m{player,encoder}.exe)
-if [[ $mplayer = y ]] && do_svn; then
+if [[ $mplayer = y ]] && check_mplayer_updates; then
     [[ $license != nonfree || $faac == n ]] && faac_opts=(--disable-faac)
     do_uninstall "${_check[@]}"
     [[ -f config.mak ]] && log "distclean" make distclean
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1556,9 +1556,6 @@ if [[ ! $x265 = n ]] && do_vcs "https://bitbucket.org/multicoreware/x265_git.git
     [[ $bits = 32bit ]] && assembly=-DENABLE_ASSEMBLY=OFF
     [[ $x265 = d ]] && xpsupport=-DWINXP_SUPPORT=ON
 
-    # Fix hg.bat using python2 still
-    grep_and_sed python2 /usr/bin/hg.bat 's/python2/python3/'
-
     build_x265() {
         create_build_dir
         local build_root=$PWD
@@ -1569,7 +1566,7 @@ if [[ ! $x265 = n ]] && do_vcs "https://bitbucket.org/multicoreware/x265_git.git
         extra_script pre cmake
         log "cmake" cmake "$(get_first_subdir -f)/source" -G Ninja \
         -DCMAKE_INSTALL_PREFIX="$LOCALDESTDIR" -DBIN_INSTALL_DIR="$LOCALDESTDIR/bin-video" \
-        -DENABLE_SHARED=OFF -DENABLE_CLI=OFF -DHIGH_BIT_DEPTH=ON -DHG_EXECUTABLE=/usr/bin/hg.bat \
+        -DENABLE_SHARED=OFF -DENABLE_CLI=OFF -DHIGH_BIT_DEPTH=ON \
         -DENABLE_HDR10_PLUS=ON $xpsupport -DCMAKE_CXX_COMPILER="$LOCALDESTDIR/bin/g++.bat" \
         -DCMAKE_TOOLCHAIN_FILE="$LOCALDESTDIR/etc/toolchain.cmake" "$@"
         extra_script post cmake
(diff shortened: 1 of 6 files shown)


ImportTaste:
Potentially unwanted issue: there's a prompt to remove mercurial, which halts the script until you confirm it.

build/media-suite_compile.sh
```diff
@@ -2054,10 +2054,71 @@ if [[ $ffmpeg != no ]]; then
     fi
 fi
 
+# static do_vcs just for svn
+do_svn() {
```
wiiaboo, line 2058: Well, might as well rename it to "check_mplayer_updates" or something.

wiiaboo:
Still can't test this myself for another 2 weeks (no internet at the home where I have a Windows).

It should relatively easy to test.

1. 32-bit and then 64-bit (if there are new commits after 32-bit was compiled, there should be a prompt or some thing that indicates 32-bit might be out-of-date so people can recompile it again if they want)
2. `patch` patches
3. `am` patches
4. patches that create new files, especially `patch` patches (they should be gone with the vcs_reset)
5. upstream force-pushes

Non-related: since we have an organization now, it could be more easy to maintain all the suite-especific patches by using a dedicated repo for them instead of trusting on 0x0.st or other 3rd party hosts.

1480c1:
> Potentially unwanted issue: there's a prompt to remove mercurial, which halts the script until you confirm it.

added
```bash
do_pacman_remove -m mercurial
```
to the update script to remove mercurial before checking

build/media-suite_compile.sh
1480c1: done

1480c1:
> Still can't test this myself for another 2 weeks (no internet at the home where I have a Windows).
> 
> It should relatively easy to test.
> 
> 1. 32-bit and then 64-bit (if there are new commits after 32-bit was compiled, there should be a prompt or some thing that indicates 32-bit might be out-of-date so people can recompile it again if they want)

testing this now just to make sure, hate powershell's rm btw

> 2. `patch` patches
> 3. `am` patches
> 4. patches that create new files, especially `patch` patches (they should be gone with the vcs_reset)

Would it be better to do a `vcs_clean` in between cloning/updating and touching the recently updated file just to remove those files? Assuming a user uses the extra scripts, they would put them in the build dir and touch any files needed so it shouldn't result in any unintended consequences

> 5. upstream force-pushes

I believe we should be fine for these since we essentially do something similar to a reset --hard upstream


> Non-related: since we have an organization now, it could be more easy to maintain all the suite-especific patches by using a dedicated repo for them instead of trusting on 0x0.st or other 3rd party hosts.

I think this would be better than using a bunch of random gists, I can work on this

build/media-suite_helper.sh
```diff
@@ -336,27 +277,27 @@ do_vcs() {
         rm -f ./build_successful{32,64}bit{,_*}
         if [[ $build32$build64$bits == yesyes64bit ]]; then
             new_updates="yes"
-            new_updates_packages="$new_updates_packages [$vcsFolder]"
+            new_updates_packages="$new_updates    oldHead=$(vcs_get_current_head "$PWD")_packages [$vcsFolder]"
```
wiiaboo, line 280: Is this correct?

1480c1: uhhh, nope I'm pretty sure

wiiaboo:
Not sure if you're doing the prompt again, it should still exist in https://github.com/m-ab-s/media-autobuild_suite/blob/master/build/media-suite_compile.sh#L2702-L2716, unless it was removed.

build/media-suite_helper.sh
1480c1: fixed, I think this was a copy paste issue

ImportTaste:
`do_pacman_remove -m mercurial` causes the 'update autobuild suite' step to soft-lock if mercurial is already removed.

1480c1:
oh, it seems I forgot to add a break condition for do_pacman_remove

ImportTaste:
I am noticing that vorbis-tools is coming up as [Recently updated] every time, which I don't think is a problem unique to this PR, but since removing mercurial un-complicates things, now is a good time to address it.

1480c1:
Do you know if it's only vorbis-tools? I'm trying to see if there's any common cross point since I can't seem to reproduce on my machine

ImportTaste:
It is only vorbis-tools.

ImportTaste:
It is only vorbis-tools, every other check came up [Up-to-date]. Just tried it with the commit you just pushed, as well.

image

ImportTaste:
There we go, that fixed it.

1480c1:
Does it print `Updated` at the end of compilation?

since the main checks for printing `Recently updated` are if old merge base != new head and `recently_updated` existing and `build_successful64bits` not existing (the main one being `recently_updated` and `build_successful$bits`). So this would trigger in the case were the repo was updated, but a build was not yet run or successful.

the `do_checkIfExist()` function is the one that is supposed to generate that file and creates it if the files it expects exists (the `_check` variable) and if `build32` or `build64` is set to yes and if the directory the repo is supposed to be in exists. This check and generation would run after printing `Updated`

ps. it seems this took a long time to type and by the time I typed this, you already fixed it.

1480c1:
```none
10:00:07 Finished 64bit compilation of all tools
There were new updates while compiling.
Updated: [SVT-HEVC] [vvc]
Would you like to run compilation again to get those updates? Default: no
y/[n]
```
prompt thingy seems to work

1480c1:
going to merge maybe by tomorrow unless something seems broken

PR #1779, opened 2020-08-07
.bat: update with newer zeranoe included libs

wiiaboo:
Not too sure about adding libvmaf as it needs w32threads disabled.

Any opinion on this?


1480c1:
I think libvmaf is okay, only issue I would have with it is the fact that you will have to use model_path if you want to use it, but other than that, no negative opinions on it

PR #1783, opened 2020-08-11
openal-soft: remove 0001-CMake-Fix-issues-for-mingw-w64.patch

LigH-de:
Doesn't apply anymore. Builds without it now.


1480c1:
Can you try again? I rebased the patch, although the only thing that changed was line numbers

LigH-de:
OK, my mistake. Without the patch, openal is not built, but an existing library file was found.

Rebased patch works.

PR #1828, opened 2020-10-06
vulkan-loader: rebase and add support for uasm as masm-like assembler

wiiaboo:
@1480c1 I don't have write-access to mabs-patches

Also, since you're probably more familiar than me with CMake, am I doing something wrong with the MASM detection part?
Do you know any reason why we shouldn't do this?

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1802,9 +1802,10 @@ _check=(libvulkan.a vulkan.pc vulkan/vulkan.h d3d{kmthk,ukmdt}.h)
 if { { [[ $ffmpeg != no ]] && enabled vulkan; } || ! mpv_disabled vulkan; } &&
     do_vcs "https://github.com/KhronosGroup/Vulkan-Loader.git" vulkan-loader; then
     _DeadSix27=https://raw.githubusercontent.com/DeadSix27/python_cross_compile_script/master
+    _mabs=https://raw.githubusercontent.com/m-ab-s/mabs-patches/master
     _shinchiro=https://raw.githubusercontent.com/shinchiro/mpv-winbuild-cmake/master
     do_uninstall "${_check[@]}"
-    do_patch "$_shinchiro/packages/vulkan-0001-cross-compile-static-linking-hacks.patch" am
+    do_patch "$_mabs/vulkan-loader/vulkan-0001-cross-compile-static-linking-hacks.patch" am
     create_build_dir
     log dependencies /usr/bin/python3 ../scripts/update_deps.py --no-build
     cd_safe Vulkan-Headers
@@ -1816,11 +1817,11 @@ if { { [[ $ffmpeg != no ]] && enabled vulkan; } || ! mpv_disabled vulkan; } &&
         do_install d3d{kmthk,ukmdt}.h include/
     cd_safe "$(get_first_subdir -f)"
     do_print_progress "Building Vulkan-Loader"
-    CFLAGS+=" -DSTRSAFE_NO_DEPRECATE" do_cmakeinstall -DBUILD_TESTS=OFF -DCMAKE_SYSTEM_NAME=Windows -DUSE_CCACHE=OFF \
+    CFLAGS+=" -DSTRSAFE_NO_DEPRECATE" do_cmakeinstall -DBUILD_TESTS=OFF -DUSE_CCACHE=OFF \
     -DCMAKE_ASM_COMPILER="$(command -v nasm.exe)" -DVULKAN_HEADERS_INSTALL_DIR="$LOCALDESTDIR" \
     -DENABLE_STATIC_LOADER=ON -DUNIX=OFF
     do_checkIfExist
-    unset _DeadSix27 _shinchiro
+    unset _DeadSix27 _mabs _shinchiro
 fi
 
 _check=(lib{glslang,OSDependent,HLSL,OGLCompiler,SPVRemapper}.a
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1438,6 +1438,7 @@ create_build_dir() {
     shift $((OPTIND - 1))
 
     build_dir="${build_root:+$build_root/}build${1:+-$1}-$bits"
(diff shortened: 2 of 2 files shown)


1480c1:
Just added you to it, didn't know you didn't have perms there

going to check in a bit for the cmake stuff

build/media-suite_compile.sh
```diff
@@ -1816,11 +1817,12 @@ if { { [[ $ffmpeg != no ]] && enabled vulkan; } || ! mpv_disabled vulkan; } &&
         do_install d3d{kmthk,ukmdt}.h include/
     cd_safe "$(get_first_subdir -f)"
     do_print_progress "Building Vulkan-Loader"
-    CFLAGS+=" -DSTRSAFE_NO_DEPRECATE" do_cmakeinstall -DBUILD_TESTS=OFF -DCMAKE_SYSTEM_NAME=Windows -DUSE_CCACHE=OFF \
-    -DCMAKE_ASM_COMPILER="$(command -v nasm.exe)" -DVULKAN_HEADERS_INSTALL_DIR="$LOCALDESTDIR" \
+    CC=gcc CXX=g++ CFLAGS+=" -DSTRSAFE_NO_DEPRECATE" do_cmakeinstall -DBUILD_TESTS=OFF \
+    -DCMAKE_SYSTEM_NAME=Windows -DUSE_CCACHE=OFF \
```
1480c1, line 1821: btw, why do we set `CMAKE_SYSTEM_NAME` here? Isn't it usually `Windows` either way?

https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html

https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_NAME.html

wiiaboo:
It's probably organization settings that only give basic read access to members.

build/media-suite_compile.sh
```diff
@@ -1816,11 +1817,12 @@ if { { [[ $ffmpeg != no ]] && enabled vulkan; } || ! mpv_disabled vulkan; } &&
         do_install d3d{kmthk,ukmdt}.h include/
     cd_safe "$(get_first_subdir -f)"
     do_print_progress "Building Vulkan-Loader"
-    CFLAGS+=" -DSTRSAFE_NO_DEPRECATE" do_cmakeinstall -DBUILD_TESTS=OFF -DCMAKE_SYSTEM_NAME=Windows -DUSE_CCACHE=OFF \
-    -DCMAKE_ASM_COMPILER="$(command -v nasm.exe)" -DVULKAN_HEADERS_INSTALL_DIR="$LOCALDESTDIR" \
+    CC=gcc CXX=g++ CFLAGS+=" -DSTRSAFE_NO_DEPRECATE" do_cmakeinstall -DBUILD_TESTS=OFF \
```
1480c1, line 1820: any reason why we are setting CC and CXX here? Is ccache or clang broken?

As for the using uasm, I think the issue might be that neither [this line](https://github.com/KhronosGroup/Vulkan-Loader/commit/e1c938170bba5879d95bf2d7b57b4d1d68557cd5#diff-ac8626da51dc3b5d11ecb20f761f7344R137) nor cmake's ASM_MASM language will pick up on it, https://github.com/Kitware/CMake/blob/master/Modules/CMakeDetermineASM_MASMCompiler.cmake

We might need to force set it through `-DCMAKE_ASM_MASM_COMPILER="$(command -v uasm)"`

wiiaboo: It was broken with the previous check using `COMMAND ${CMAKE_C_COMPILER} --version OUTPUT_VARIABLE COMPILER_VERSION_OUTPUT` but I refactored it to use `CMAKE_SIZEOF_VOID_P STREQUAL "8"` instead, so it's probably not needed anymore. I think CMake was assuming the `ccache` part of `C=ccache gcc` as the C compiler.

It detects uasm fine and even compiles fine. It just can't identify it as masm. But I think both jwasm and uasm were made to be plug-and-play replacements of masm.

build/media-suite_compile.sh
wiiaboo: No idea, probably cargo-culted from somewhere.

wiiaboo:
Welp, nevermind, doesn't seem to work.

1480c1:
Seems to also be broken for 32-bit builds as well
```c
[10/14] Building ASM_MASM object loader/CMakeFiles/loader-unknown-chain.dir/unknown_ext_chain_masm.asm.obj
FAILED: loader/CMakeFiles/loader-unknown-chain.dir/unknown_ext_chain_masm.asm.obj 
D:\mabs\msys64\mingw32\bin\uasm.exe -DAPI_NAME=\"Vulkan\" -DVK_ENABLE_BETA_EXTENSIONS -DVK_USE_PLATFORM_WIN32_KHR -DWIN32_LEAN_AND_MEAN -I../loader -I../loader/generated -Iloader /coff /c  /Fo loader/CMakeFiles/loader-unknown-chain.dir/unknown_ext_chain_masm.asm.obj ../loader/unknown_ext_chain_masm.asm
UASM v2.50, Jan  8 2020, Masm-compatible assembler.
Portions Copyright (c) 1992-2002 Sybase, Inc. All Rights Reserved.
Source code is available under the Sybase Open Watcom Public License.

../loader/unknown_ext_chain_masm.asm(115) : Error A2030: Instruction or register not accepted in current CPU mode
../loader/unknown_ext_chain_masm.asm(120) : Error A2082: Must be in segment block
../loader/unknown_ext_chain_masm.asm(130) : Error A2082: Must be in segment block
 PhysDevExtTramp(2)[unknown_ext_chain_masm.asm]: Macro called from
  ../loader/unknown_ext_chain_masm.asm(130): Main line code
../loader/unknown_ext_chain_masm.asm(130) : Error A2082: Must be in segment block
...
```

wiiaboo:
Well, this was an exercise in pointlessness.

1480c1:
Well, compiling this succeeded for both 32-bit and 64-bit and gcc and clang. No idea how to quickly test it though

1480c1:
@wiiaboo do you want to merge?

PR #1831, opened 2020-10-09
xvid_encraw: replace sed with zones+stdin patch

### LINKED ISSUE #1829: Request: Enable stdin as default input in xvid_encraw
The help output of **xvid_encraw** promises that the default input is **stdin** if the `-i` parameter is omitted. But the source is incomplete, it does not work this way when compiled under Windows. Was discussed in the [doom9 forum](https://forum.doom9.org/showthread.php?t=181894) by @Selur – and @Kurtnoise published a patched version.

Here is a diff patch comparing Kurtnoise's source with the official xvid 1.3.7, including the 8192 zones `sed` which was applied in **media-suite_compile.sh** already (and a few spacing cosmetics which happened while editing).

```
Index: xvidcore/examples/xvid_encraw.c
===================================================================
--- xvidcore/examples/xvid_encraw.c	(revision 2196)
+++ xvidcore/examples/xvid_encraw.c	(working copy)
@@ -38,6 +38,16 @@
  *  Use ./xvid_encraw -help for a list of options
  *	
  ************************************************************************/
+#ifdef _WIN32
+/* The following two defines must be located before the inclusion of any system header files. */
+#define WINVER       0x0500
+#define _WIN32_WINNT 0x0500
+#include <windows.h>
+#include <io.h>       /* _setmode() */
+#include <fcntl.h>    /* _O_BINARY */
+#endif
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -124,7 +134,7 @@
  *                     Command line global variables
  ****************************************************************************/
 
-#define MAX_ZONES   64
+#define MAX_ZONES   8192
 #define MAX_ENC_INSTANCES 4
 #define MAX_XDIM 4096
 #define MAX_YDIM 4096
@@ -350,6 +360,7 @@
 	double totalenctime = 0.;
 	float totalPSNR[3] = {0., 0., 0.};
 
+	FILE *in_file = stdin;
 	FILE *statsfile;
 	frame_stats_t framestats[7];
 
@@ -990,8 +1001,14 @@
 	  	  if (get_frame) AVIStreamGetFrameClose(get_frame);
 		  if (avi_in_stream) AVIStreamRelease(avi_in_stream);
 		  AVIFileExit();
-      }
-      else
+		}
+		else if (ARG_INPUTFILE == NULL || strcmp(ARG_INPUTFILE, "stdin") == 0) {
+#ifdef _WIN32
+		  _setmode(_fileno(stdin), _O_BINARY);
+#endif
+		  in_file = stdin;
+		}
+		else
 #endif
 		{
 			FILE *in_file = fopen(ARG_INPUTFILE, "rb");
```

[xvid_encraw.c.mabs.zip](https://github.com/m-ab-s/media-autobuild_suite/files/5346538/xvid_encraw.c.mabs.zip)

LigH-de:
replace sed which patched MAX_ZONES only with a patch by Kurtnoise that also enables stdin as default input under Windows;
you may want to move this patch to your own repo instead of relying on my server

Fixes #1829

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1664,7 +1664,7 @@ if enabled libxvid && [[ $standalone = y ]] && ! { files_exist "${_check[@]}" &&
     do_install '=build/xvidcore.a' libxvidcore.a
     do_install '=build/xvidcore.dll' bin-video/
     cd_safe ../../examples
-    sed -ri "s;(#define MAX_ZONES\s*) \S.*$;\1 8192;" xvid_encraw.c
+    do_patch "https://www.ligh.de/software/xvid_encraw.c.mabs.patch"
     do_make xvid_encraw
     do_install xvid_encraw.exe bin-video/
     do_checkIfExist


wiiaboo:
Can you try with this patch instead?:
```patch
diff -rwU3 xvidcore/examples/xvid_encraw.c xvidcore_patched/examples/xvid_encraw.c
--- xvidcore/examples/xvid_encraw.c	2019-12-28 14:19:24.000000000 +0000
+++ xvidcore_patched/examples/xvid_encraw.c	2020-10-09 16:39:35.000000000 +0100
@@ -47,7 +47,10 @@
 #ifndef WIN32
 #include <sys/time.h>
 #else
+#define WINVER       0x0500
+#define _WIN32_WINNT 0x0500
 #include <windows.h>
+#include <fcntl.h>    /* _setmode() (thru io.h) and _O_BINARY */
 #include <vfw.h>
 #include <time.h>
 #define XVID_AVI_INPUT
@@ -1313,6 +1316,9 @@
 #endif
 
 	if (ARG_INPUTFILE == NULL || strcmp(ARG_INPUTFILE, "stdin") == 0) {
+#ifdef _WIN32
+		_setmode(_fileno(stdin), _O_BINARY);
+#endif
 		in_file = stdin;
 	} else {
 #ifdef XVID_AVI_INPUT
```

You'll need to apply it with `-p2` if you want to apply it on the same directory.

With this you shouldn't need `-i stdin` (and it probably won't work, as the ~999L code is still going to open "stdin" as "rb"). Just don't use `-i`.

1480c1:
@wiiaboo with your patch
```powershell
ffmpeg -i D:\cone.y4m -f yuv4mpegpipe - | xvid_encraw.exe -o temp.xvid
```
works

wiiaboo:
I've never used encraw, even in the olden times. I would expect for stdin input that it either autodetects what's coming in or there's options to force it to choose one format or the other, like x264.
Can't you force it to interpret stdin as y4m?

1480c1:
```none
ffmpeg -i D:\cone.y4m -f rawvideo - | xvid_encraw.exe -o temp.xvid -framerate 29.97 -w 1080 -h 1680
```
seems to work

from what I could see in the help output
```none
 -type   integer : input data type (yuv=0, pgm=1, avi/avs=2) (0)
```
it seems only yuv, pgm, or avi/avs is accepted, no y4m

LigH-de:
AFAIK, xvid_encraw does not contain a Y4M header parser. As I wrote somewhere else: It seems to be rather only a test tool than a production application. But it exhibits some parameters ffmpeg may not support well.

1480c1:
btw, line I used was

```bash
do_patch "https://github.com/m-ab-s/xvid/compare/lighde.patch" "" 2
```

1480c1:
image
I think I got a parameter wrong

LigH-de:
You will need CSP i420 instead of yv12.

1480c1:
seems it was set to yv12 instead of i420 by default
image

LigH-de:
Ongoing discussions: [doom9 forum](http://forum.doom9.org/showthread.php?t=181894) and [VideoHelp forum](https://forum.videohelp.com/threads/399070)

wiiaboo:
So basically, @LigH-de, what else is missing? Do you really need `-i stdin` to work? The `ARG_INPUTFILE == NULL` part in Kurtnoise's patch does nothing because that is inside a `if (ARG_INPUTFILE != NULL)`.

wiiaboo:
Other than the ZONES mod, of course, which will still be done through `sed`.

wiiaboo:
If you're going to submit it to xvid, I'd remove the `_WIN32_WINNT` and `WINVER` crap. Better not encourage crazy idiots still on ancient OSes.

I'm fine with not getting a mention, I just grabbed the previous patch and moved stuff I thought didn't make sense. Didn't even test it.

1480c1:
btw, seems I did the git svn wrong, so I might need to force push that repository.

LigH-de:
I do not need `-i stdin`; I said that "no `-i` parameter" did not work in my first test. If it works now, that will be fine (for @Selur too), then `-i stdin` as backup is not required.

1480c1:
Okay, once I finally figure out the "ideal" way of doing an svn->git mirror, I will change the do_wget to a do_vcs with the mirror and apply the patch onto that

LigH-de:
I tested it again... with my original patch based on Kurtnoise's source, xvid_encraw aborts prematurely with omitted `-i` parameter. But the whole input is converted reliably with the `-i stdin` parameter. So there must have been a difference in the file handling.

I will now try your edited patch.

wiiaboo:
@LigH-de at least from my interpretation of the code, I don't think `-i stdin` ever worked. Or maybe "stdin" is especially handled by `fopen`.
There are some ifs for the case ARG_INPUTFILE is "stdin", though, so it either that just works in *nix, or they forgot to add a `else if (strcmp(ARG_INPUTFILE, "stdin") == 0`)` before the `else` inside `if (ARG_INPUTFILE != NULL)`.
Either way, if `-i stdin` is supposed to **officially** work, I think it's very zero-effort to add it for Windows.

What I assume from the help's text is «if `-i` is not used, it's assumed "stdin"», not that "-i stdin" should work.

LigH-de:
I agree, the `stdin` in the help text was meant as symbolic, not literal statement. The case of the absent input parameter is relevant.

1480c1:
Is
```c
	if (ARG_INPUTFILE == NULL || strcmp(ARG_INPUTFILE, "stdin") == 0) {
		in_file = stdin;
	}
```
in `encode_sequence` not supposed to handle it by catching when you don't specify `-i` or anything to `-i` and if the value you set to `-i` is a string that matches `stdin`?

wiiaboo:
Sure, except the first `if` inside `if (ARG_INPUTFILE != null)`  in `main()` is for avs/avi input and then there's only an `else` that tries to `fopen` the file and seek to the end.
Like I said, it's missing an `else if` for the `-i stdin` case.

PR #1895, opened 2020-12-24
Fix avif & jpegxl

Jamaika1:

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1091,14 +1091,14 @@ fi
 _check=(libavif.{a,pc} avif/avif.h)
 [[ $standalone = y ]] && _check+=(bin-video/avif{enc,dec}.exe)
 if [[ $libavif = y ]] && {
-        pc_exists "aom" || pc_exists "dav1d" || pc_exists "rav1e" || pc_exists "libgav1"
+        pc_exists "aom" || pc_exists "dav1d" || pc_exists "rav1e" || pc_exists "svtav1"
     } &&
     do_vcs "https://github.com/AOMediaCodec/libavif.git"; then
     do_uninstall "${_check[@]}"
     do_pacman_install libjpeg-turbo
     do_patch "https://raw.githubusercontent.com/m-ab-s/mabs-patches/master/libavif/0001-CMake-Use-the-import-libraries-and-the-proper-variab.patch" am
     extracommands=()
-    pc_exists "libgav1" && extracommands+=("-DAVIF_CODEC_LIBGAV1=ON")
+    pc_exists "svtav1" && extracommands+=("-DAVIF_CODEC_SVT=ON")
     pc_exists "dav1d" && extracommands+=("-DAVIF_CODEC_DAV1D=ON")
     pc_exists "rav1e" && extracommands+=("-DAVIF_CODEC_RAV1E=ON")
     pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON")
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1101,7 +1101,7 @@ if [[ $libavif = y ]] && {
     pc_exists "svtav1" && extracommands+=("-DAVIF_CODEC_SVT=ON")
     pc_exists "dav1d" && extracommands+=("-DAVIF_CODEC_DAV1D=ON")
     pc_exists "rav1e" && extracommands+=("-DAVIF_CODEC_RAV1E=ON")
-    pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON")
+    pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON -DAVIF_CODEC_AOM_DECODE=OFF -DAVIF_CODEC_AOM_ENCODE=ON")
     case $standalone in
     y) extracommands+=("-DAVIF_BUILD_APPS=ON") ;;
     *) extracommands+=("-DAVIF_BUILD_APPS=OFF") ;;
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1091,16 +1091,17 @@ fi
 _check=(libavif.{a,pc} avif/avif.h)
 [[ $standalone = y ]] && _check+=(bin-video/avif{enc,dec}.exe)
 if [[ $libavif = y ]] && {
-        pc_exists "aom" || pc_exists "dav1d" || pc_exists "rav1e"
+        pc_exists "aom" || pc_exists "dav1d" || pc_exists "rav1e" || pc_exists "libgav1"
     } &&
     do_vcs "https://github.com/AOMediaCodec/libavif.git"; then
     do_uninstall "${_check[@]}"
     do_pacman_install libjpeg-turbo
     do_patch "https://raw.githubusercontent.com/m-ab-s/mabs-patches/master/libavif/0001-CMake-Use-the-import-libraries-and-the-proper-variab.patch" am
     extracommands=()
+    pc_exists "libgav1" && extracommands+=("-DAVIF_CODEC_LIBGAV1=ON")
     pc_exists "dav1d" && extracommands+=("-DAVIF_CODEC_DAV1D=ON")
     pc_exists "rav1e" && extracommands+=("-DAVIF_CODEC_RAV1E=ON")
-    pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON")
+    pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON")
     case $standalone in
     y) extracommands+=("-DAVIF_BUILD_APPS=ON") ;;
     *) extracommands+=("-DAVIF_BUILD_APPS=OFF") ;;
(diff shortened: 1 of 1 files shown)


1480c1:
Can you describe what issue exactly is it fixing? your commit is a bit vague as I am not aware of any specific issue with those two other than jpeg-xl can't use simd with gcc

build/media-suite_compile.sh
```diff
@@ -1091,16 +1091,17 @@ fi
...
     do_vcs "https://github.com/AOMediaCodec/libavif.git"; then
     do_uninstall "${_check[@]}"
     do_pacman_install libjpeg-turbo
     do_patch "https://raw.githubusercontent.com/m-ab-s/mabs-patches/master/libavif/0001-CMake-Use-the-import-libraries-and-the-proper-variab.patch" am
     extracommands=()
+    pc_exists "libgav1" && extracommands+=("-DAVIF_CODEC_LIBGAV1=ON")
     pc_exists "dav1d" && extracommands+=("-DAVIF_CODEC_DAV1D=ON")
     pc_exists "rav1e" && extracommands+=("-DAVIF_CODEC_RAV1E=ON")
-    pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON")
+    pc_exists "aom" && extracommands+=("-DAVIF_CODEC_AOM=ON -DAVIF_CODEC_AOM_DECODE=ON -DAVIF_CODEC_AOM_ENCODE=ON")
```
wiiaboo, line 1104: What does this change?

build/media-suite_compile.sh
```diff
@@ -1113,6 +1114,7 @@ _check=(bin-global/{{c,d}j,benchmark_}xl.exe)
 if [[ $jpegxl = y ]] && do_vcs "https://gitlab.com/wg1/jpeg-xl.git"; then
     do_uninstall "${_check[@]}"
     do_pacman_install lcms2
+    do_pacman_install libjpeg-turbo
```
wiiaboo, line 1117: libjpeg-turbo I think is a dependency of lcms2, or vice-versa. No need for both.
Or if both are indeed needed you can just call do_pacman_install once with both parameters.

build/media-suite_compile.sh
```diff
@@ -1121,7 +1123,7 @@ if [[ $jpegxl = y ]] && do_vcs "https://gitlab.com/wg1/jpeg-xl.git"; then
     fi
     CXXFLAGS+=" ${extra_cxxflags[*]}" \
         do_cmake global -D{BUILD_TESTING,JPEGXL_{ENABLE_OPENEXR,ENABLE_SKCMS}}=OFF \
-        -DJPEGXL_{FORCE_SYSTEM_BROTLI,STATIC,ENABLE_BENCHMARK}=ON
+        -DJPEGXL_{FORCE_SYSTEM_HWY,FORCE_SYSTEM_BROTLI,STATIC,ENABLE_BENCHMARK}=ON
```
wiiaboo, line 1126: What does this do?

wiiaboo:
And the commit message is indeed poor. What does this fix and why?

build/media-suite_compile.sh
Jamaika1: Google highway was originally inside the jpegxl codec on a third party. Now it is separate library on github.
https://github.com/google/highway

build/media-suite_compile.sh
Jamaika1: I don't know what is the creator's vision. They see some libraries are in pacman. Other libraries such as libtiff, libpng aren't.

build/media-suite_compile.sh
Jamaika1: In GCC manually adding the AVIF_CODEC_AOM definition doesn't activate AVIF_CODEC_AOM_DECODE and AVIF_CODEC_AOM_ENCODE. Unless they are to be inactive.

Jamaika1:
> Can you describe what issue exactly is it fixing? your commit is a bit vague as I am not aware of any specific issue with those two other than jpeg-xl can't use simd with gcc

No problem just my curiosity about how add-ons work. I don't even know what codec versions you are adding.

build/media-suite_compile.sh
1480c1: that might be related to a pkgconf bug I think

Jamaika1:
As for libavif. Today there is libheif from libavif on the market and google libwebp2 product from av1enc which has libavif. It is difficult to say which library is more advanced.
My question is also which decoder for aom is better dav1d or libgav1?
Dav1d doesn't work with gcc for 16bit for aom 10 / 12bit HDR.
libgav1 is only 8.10 bits.

build/media-suite_compile.sh
Jamaika1: Why use lcms2 if there is google skcms in c++11?

Jamaika1:
> And the commit message is indeed poor. What does this fix and why?
It's not so much about bugs, but about newer versions of codecs and their possibilities.

build/media-suite_compile.sh
wiiaboo: Nope, like I thought, this doesn't do shit. Remove this line's changes.

You'd only use these options if you want to **selectively** enable/disable encoding **or** decoding with AOM, if AOM is enabled. As in, if you're enabling both dav1d/gav1 (i.e. good/fast decoders) and aom, you wouldn't use aom for decoding, so there's no point including it.

build/media-suite_compile.sh
wiiaboo: Because mpv already uses lcms2.
Collapse these two into the same line. They're not dependencies.

build/media-suite_compile.sh
wiiaboo: And where exactly is highway being installed/compiled?
I see it still as a git submodule in jpeg-xl's repo.

Remove this change.

wiiaboo:
> but about newer versions of codecs and their possibilities.

Other than adding libgav1, without adding how to compile/install it, I don't really see anything new or fixed.

build/media-suite_compile.sh
Jamaika1: with JPEGXL 0.2.0
```
# Highway
set(HWY_SYSTEM_GTEST ON CACHE INTERNAL "")
if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/highway/CMakeLists.txt" AND
    NOT JPEGXL_FORCE_SYSTEM_HWY)
  add_subdirectory(highway)
  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/highway/LICENSE"
                 ${PROJECT_BINARY_DIR}/LICENSE.highway COPYONLY)
else()
  pkg_check_modules(HWY libhwy)
  if (NOT HWY_FOUND)
    message(FATAL_ERROR
        "Highway library (hwy) not found. Install libhwy-dev or download it "
        "to third_party/highway from https://github.com/google/highway . "
        "Highway is required to build JPEG XL. You can run "
        "${PROJECT_SOURCE_DIR}/deps.sh to download this dependency.")
  endif()
  add_library(hwy INTERFACE IMPORTED GLOBAL)
  if(${CMAKE_VERSION} VERSION_LESS "3.13.5")
    set_property(TARGET hwy PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${HWY_INCLUDE_DIR})
    target_link_libraries(hwy INTERFACE ${HWY_LDFLAGS})
    set_property(TARGET hwy PROPERTY INTERFACE_COMPILE_OPTIONS ${HWY_CFLAGS_OTHER})
  else()
    target_include_directories(hwy INTERFACE ${HWY_INCLUDE_DIRS})
    target_link_libraries(hwy INTERFACE ${HWY_LINK_LIBRARIES})
    target_link_options(hwy INTERFACE ${HWY_LDFLAGS_OTHER})
    target_compile_options(hwy INTERFACE ${HWY_CFLAGS_OTHER})
  endif()
  if(JPEGXL_DEP_LICENSE_DIR)
    configure_file("${JPEGXL_DEP_LICENSE_DIR}/libhwy-dev/copyright"
                   ${PROJECT_BINARY_DIR}/LICENSE.highway COPYONLY)
  endif()  # JPEGXL_DEP_LICENSE_DIR
endif()
```

build/media-suite_compile.sh
Jamaika1: I understand. skcms is not included in mpv, there is no need to use the library.

build/media-suite_compile.sh
Jamaika1: If dav1d/libgav1 is on then it will be understood that AVIF_CODEC_AOM_DECODE must be off. What do we do with AVIF_CODEC_AOM_ENCODE encoder? Also redundant function?
I am surprised that AOM decoder doesn't work with assembler and needs other decoders for amateur purposes. GCC itself is known not to be suitable.

Jamaika1: #error AVIF_CODEC_AOM_DECODE or AVIF_CODEC_AOM_ENCODE must be defined
```
#if defined(AVIF_CODEC_AOM)
    { AVIF_CODEC_CHOICE_AOM,
      "aom",
      avifCodecVersionAOM,
      avifCodecCreateAOM,
#if defined(AVIF_CODEC_AOM_DECODE) && defined(AVIF_CODEC_AOM_ENCODE)
      AVIF_CODEC_FLAG_CAN_DECODE | AVIF_CODEC_FLAG_CAN_ENCODE
#elif defined(AVIF_CODEC_AOM_DECODE)
      AVIF_CODEC_FLAG_CAN_DECODE
#elif defined(AVIF_CODEC_AOM_ENCODE)
      AVIF_CODEC_FLAG_CAN_ENCODE
#else
#error AVIF_CODEC_AOM_DECODE or AVIF_CODEC_AOM_ENCODE must be defined
#endif
```

wiiaboo: One of them must be defined and enabled _if_ AVIF_CODEC_AOM is enabled.
So like I said, there's no need to explicitely use them in the suite. If people really want to exclude just the decoder or just the encoder, they can use [Custom Patches](https://github.com/m-ab-s/media-autobuild_suite#custom-patches) to add the extra cmake options if they want.

Remove this change.

build/media-suite_compile.sh
wiiaboo: I fail to see how that refutes what I said.
We don't need highway for anything else, and it's not available as a MINGW package, so there's no point in using the system's highway.

PR #1969, opened 2021-05-02
update SVT-AV1 repo URL

Patman86:
github is archived, only gitlab is used
you know... 😉😜


1480c1:
kek, forgot to update it

PR #2102, opened 2022-01-19
update python and vapoursynth

woot000:
it builds fine and seems to work in ffmpeg


1480c1:
Do you happen to have Windows 7 machine to test compilation and also running it?

Python 3.9 officially dropped support for Win7 and has moved to using some APIs from 8.1 and on, but msys2 has some patches, so their python still works with win7, but I don't know if vapoursynth's stuff still works with win7

woot000:
unfortunately not
i'd test it on a virtual machine but vmware has been acting buggy for me on my windows 11 machine so i wouldn't be able to say for sure that vapoursynth would work

1480c1:
🤷

1480c1:
Thanks for the help

Issue #2100, opened 2022-01-10
[mingw64] [libjxl] failed to link with gflags


Doofussy2:
Failed build with cmake package.

```
Likely error (tail of the failed operation logfile):
  but not all the files it references.

Call Stack (most recent call first):
  G:/mpv_building_tools/msys64/mingw64/lib/cmake/gflags/gflags-config.cmake:17 (include)
  third_party/CMakeLists.txt:168 (find_package)
```
I am building 64bit on Windows 10. I initially was getting this error, and tried to resolve it. I tried replacing the `gflags` with the latest. On that failing, I rebuilt the entire suite. The error persists.

[logs.zip](https://0x0.st/oieV.zip)

1480c1:
Seems it occurred when msys2 added the shared library to the package, resulting in the cmake file referring to the dll import library, that we locally move so gcc doesn't find it, building locally to see what would be the best way to move forward.

LigH-de:
Excerpt from libjxl:
```
CMake Error at G:/MABS/msys64/mingw32/lib/cmake/gflags/gflags-nonamespace-targets.cmake:107 (message):
  The imported target "gflags_shared" references the file

     "G:/MABS/msys64/mingw32/lib/libgflags.dll.a"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "G:/MABS/msys64/mingw32/lib/cmake/gflags/gflags-nonamespace-targets.cmake"

  but not all the files it references.
```

The file `G:/MABS/msys64/mingw32/lib/libgflags.dll.a` does not exit.
A file  `G:/MABS/msys64/mingw32/lib/libgflags.dll.a.dyn` does exit instead.

bombenbodo:
I‘m getting the same error as above, but libgflags.dll.a actually exists in /msys64/mingw32/lib/

LigH-de:
Did you ever build it more than 3 weeks ago? Maybe it's an older remnant. I set up a completely new suite before I got this error in my case.

bombenbodo:
Yes, my last build/update was back in september. 

Also strange is:
```
libjxl git  ......................................... [Recently updated]
├ Uninstalling asciidoc-py3-git... failed
```

LigH-de:
This attempt fails when it has been removed successfully before. May be ignored.

[...]

PR #2106, opened 2022-01-22
jpeg-xl: fix cmake configuration

woot000:
fixes m-ab-s/media-autobuild_suite#2100


1480c1:
looks good enough to me, won't be able to test it since my laptop will be off for a while

1480c1:
Thanks

PR #2196, opened 2022-05-24
using correct name for VVC software

lehmann-c:
- misleading  VVC software name (Fraunhofer VVC), which is not used in the build script
- build script is using the VVC Reference Software VTM (https://gitlab.com/m-ab-s/VVCSoftware_VTM.git) 
- if Fraunhofer VVC software should be used, VVenC/VVdeC have to be added as a build step


1480c1:
Thanks

Issue #1535, opened 2020-01-21
FFmpeg fails with libvmaf enabled : undefined reference to `fatal'


pat357:
As  title says.

Logs at https://0x0.st/zCS8.zip

``` bash
H:/Mingw645/build/ffmpeg-git/fftools/cmdutils.c: In function 'print_buildconf':
H:/Mingw645/build/ffmpeg-git/fftools/cmdutils.c:1170:9: warning: 'strncpy' output truncated before terminating nul copying 3 bytes from a string of the same length [-Wstringop-truncation]
 1170 |         strncpy(conflist, "~--", 3);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
H:/Mingw645/build/ffmpeg-git/fftools/cmdutils.c:1176:9: warning: 'strncpy' output truncated before terminating nul copying 11 bytes from a string of the same length [-Wstringop-truncation]
 1176 |         strncpy(remove_tilde, "pkg-config ", 11);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
AR      libswscale/libswscale.a
H:/Mingw645/build/ffmpeg-git/fftools/ffmpeg.c: In function 'init_output_stream':
H:/Mingw645/build/ffmpeg-git/fftools/ffmpeg.c:3533:9: warning: 'avcodec_copy_context' is deprecated [-Wdeprecated-declarations]
 3533 |         ret = avcodec_copy_context(ost->st->codec, ost->enc_ctx);
      |         ^~~
In file included from H:/Mingw645/build/ffmpeg-git/libavformat/avformat.h:317,
                 from H:/Mingw645/build/ffmpeg-git/fftools/ffmpeg.c:43:
H:/Mingw645/build/ffmpeg-git/libavcodec/avcodec.h:4259:5: note: declared here
 4259 | int avcodec_copy_context(AVCodecContext *dest, const AVCodecContext *src);
      |     ^~~~~~~~~~~~~~~~~~~~
H:/Mingw645/build/ffmpeg-git/fftools/ffmpeg.c:3533:9: warning: 'codec' is deprecated [-Wdeprecated-declarations]
 3533 |         ret = avcodec_copy_context(ost->st->codec, ost->enc_ctx);
      |         ^~~
In file included from H:/Mingw645/build/ffmpeg-git/fftools/ffmpeg.c:43:
[...]

PR #2221, opened 2022-06-23
Update media-suite_compile.sh

ls-milkyway:
Description
VMAF support for AOM

Add VMAF support to aomedia so that command `--tune=vmaf` can be used with standalone aom encoder.

As mentioned in the [AOM's  readme.md](https://aomedia.googlesource.com/aom/#build-with-vmaf)
I have compiled & successfully used `--tune=vmaf_with_preprocessing` with the standalone aomenc encoder built by the script.

Please note that the default VMAF model (“/usr/local/share/model/vmaf_v0.6.1.json”) will be used unless you set the following flag when running the encoder:
`--vmaf-model-path=path/to/model`


1480c1:
Can you confirm that the issues mentioned in https://github.com/m-ab-s/media-autobuild_suite/issues/1535 does not appear anymore?

ls-milkyway:
> Can you confirm that the issues mentioned in #1535 does not appear anymore?

I have successfully built with commit `2da55ddb75cfd009bb4b526f5cc8a6b15f1a860a` on Jun 16, 2022 with no problems whatsoever ...even with vvenc & vvdec ...using non-free license ......the issue https://github.com/m-ab-s/media-autobuild_suite/issues/1535 is very old .....do you mean it still persists? OR is it appearing again with my pull request??

1480c1:
> is it appearing again with my pull request in other license or settings??

This one as I previously tried enabling vmaf for aom, but when aom was linked to ffmpeg, it failed at link time. I don't remember if I managed to get it fixed or not

I do see a commit for it that came after the issue, but I don't remember testing it afterwards https://github.com/m-ab-s/aom/commit/0d36ee380b206bd9b611f16e2145969e8f76a023

build/media-suite_compile.sh
```diff
@@ -1019,7 +1019,7 @@ if { [[ $aom = y ]] || [[ $libavif = y ]] || { [[ $ffmpeg != no ]] && enabled li
     do_uninstall include/aom "${_check[@]}"
     get_external_opts extracommands
     do_cmakeinstall video -DENABLE_{DOCS,TOOLS,TEST{S,DATA}}=off \
-        -DENABLE_NASM=on -DFORCE_HIGHBITDEPTH_DECODING=0 "${extracommands[@]}"
+        -DCONFIG_TUNE_VMAF=1 -DENABLE_NASM=on -DFORCE_HIGHBITDEPTH_DECODING=0 "${extracommands[@]}"
```
1480c1, line 1022: you may want to make this a conditional, based on the same conditions that libvmaf is enabled (`[[ $ffmpeg != no ]] && enabled libvmaf `) since there's no guarantee that a person enabled aom will always have vmaf enabled

ls-milkyway: I guess you are not able to get my point ....I will try again:

Well, AOM is one of the best AV1 encoders with many options/parameters to get the best results .... `--tune=` is one of them ...with total 7 possible options ...with `--tune=psnr` & `--tune=ssim` inbuilt .....others are:
`--tune=vmaf_without_preprocessing` `--tune=vmaf_with_preprocessing` ....there is also one of the best option i.e. `--tune=butteraugli`

You want to ask user an option to built AOM with vmaf in `media-autobuild_suite.bat` with it's corresponding code in `media-suite_compile.sh`? ...if that's the case then what about  `--tune=butteraugli`?
 ......**It's true that not everyone will use vmaf or butteraugli ...but I feel a comprehensive encoder like AOM should be built with all possible options** .....screenshot of my built AOM running with `--tune=vmaf_without_preprocessing`

aomenc-vmaf

gitoss: > ....there is also one of the best option i.e. `--tune=butteraugli`

For tune vmaf there is av1an or ab-av1 - not only with aom, but a bunch of encoders.

However butteraugli has been lost, it would be nice to have a way to enable it because it's useful for avif (libavif) encoding, too. Currently it requires an older libjxl lib (0.8.2) - I had tried to git this 0.8.x branch but as a mediasuite newbie ran into numerous problems while trying to compile.

There are some tickets on libjxl and libavif - it would be great to enable butteraugili for aom even with current libjxl versions, even if if only for 8 bpp.

PR #2312, opened 2022-10-26
Allow specifying a custom ffmpeg source repo

softworkz:
This patch allows to specify a path to a custom ffmpeg repo in the `media_autobuild_suite.ini` file:

image

It can be a URL or a local path.

### DIFF
--- a/media-autobuild_suite.bat
+++ b/media-autobuild_suite.bat
@@ -149,7 +149,7 @@ set iniOptions=arch license2 vpx2 x2643 x2652 other265 flac fdkaac mediainfo ^
 soxB ffmpegB2 ffmpegUpdate ffmpegChoice mp4box rtmpdump mplayer2 mpv cores deleteSource ^
 strip pack logging bmx standalone updateSuite aom faac exhale ffmbc curl cyanrip2 ^
 rav1e ripgrep dav1d libavif vvc uvg266 jq dssim avs2 timeStamp noMintty ccache ^
-svthevc svtav1 svtvp9 xvc jo vlc CC jpegxl autouploadlogs vvenc vvdec
+svthevc svtav1 svtvp9 xvc jo vlc CC jpegxl autouploadlogs vvenc vvdec ffmpegPath
 
 set deleteIni=0
 set ini=%build%\media-autobuild_suite.ini
@@ -788,6 +788,49 @@ if %buildffmpeg%==5 set "ffmpeg=sharedlibs"
 if %buildffmpeg%==6 set "ffmpeg=bothstatic"
 if %buildffmpeg% GTR 6 GOTO ffmpeg
 if %deleteINI%==1 echo.ffmpegB2=^%buildffmpeg%>>%ini%
+
+set defaultFFmpegPath=https://git.ffmpeg.org/ffmpeg.git
+
+:ffmpegPath
+if %ffmpegPathINI%==0 (
+    set ffmpegPath=%defaultFFmpegPath%
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+    echo.
+    echo. Using default ffmpeg source path: https://git.ffmpeg.org/ffmpeg.git
+    echo.
+    echo. If you want to use a custom source repository, add a line like this 
+    echo. to media-autobuild_suite.ini:
+    echo.
+    echo.     ffmpegPath=https://github.com/username/FFmpeg.git#branch=branchname
+    echo. 
+    echo. or for a local repository like:
+    echo.
+    echo.     ffmpegPath=../myrepos/ffmpeg
+    echo.
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+) else set ffmpegPath=%ffmpegPathINI%
+
+if %deleteINI%==1 echo.ffmpegPath=%ffmpegPath%>>%ini%
+
+rem Handle relative paths and convert to absolute path
+rem after sanitizing: back- to forward-slashes, remove colon after drive letter
+call :resolvePath %ffmpegPath%
+setlocal EnableDelayedExpansion
+if exist %resolvePath% (
+    set nixdir=!resolvePath:\=/!
+    set "ffmpegPath=/!nixdir::=!"
+)
+endlocal & set "ffmpegPath=%ffmpegPath%"
+
+if not %ffmpegPath%==%defaultFFmpegPath% (
+    echo -------------------------------------------------------------------------------
+    echo.
+    echo. Using ffmpeg path: %ffmpegPath%
+    echo.
+    echo -------------------------------------------------------------------------------
+)
 
 :ffmpegUp
 if %ffmpegUpdateINI%==0 (
@@ -1751,7 +1794,7 @@ set compileArgs=--cpuCount=%cpuCount% --build32=%build32% --build64=%build64% ^
 --vvc=%vvc% --uvg266=%uvg266% --vvenc=%vvenc% --vvdec=%vvdec% --jq=%jq% --jo=%jo% --dssim=%dssim% ^
 --avs2=%avs2% --timeStamp=%timeStamp% --noMintty=%noMintty% --ccache=%ccache% --svthevc=%svthevc% ^
 --svtav1=%svtav1% --svtvp9=%svtvp9% --xvc=%xvc% --vlc=%vlc% --libavif=%libavif% --jpegxl=%jpegxl% ^
---autouploadlogs=%autouploadlogs%
+--autouploadlogs=%autouploadlogs% --ffmpegPath=%ffmpegPath% 
     set "noMintty=%noMintty%"
     if %build64%==yes ( set "MSYSTEM=MINGW64" ) else set "MSYSTEM=MINGW32"
     set "MSYS2_PATH_TYPE=inherit"
@@ -1929,4 +1972,8 @@ if not exist %instdir%\msys64\mingw%1\bin\gcc.exe (
     exit
 )
 endlocal
+goto :EOF
+
+:resolvePath
+set "resolvePath=%~dpnx1"
 goto :EOF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -35,6 +35,7 @@ while true; do
     --sox=* ) sox=${1#*=} && shift ;;
     --ffmpeg=* ) ffmpeg=${1#*=} && shift ;;
     --ffmpegUpdate=* ) ffmpegUpdate=${1#*=} && shift ;;
+    --ffmpegPath=* ) ffmpegPath="${1#*=}"; shift ;;
     --ffmpegChoice=* ) ffmpegChoice=${1#*=} && shift ;;
     --mplayer=* ) mplayer=${1#*=} && shift ;;
     --mpv=* ) mpv=${1#*=} && shift ;;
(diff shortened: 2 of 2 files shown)


media-autobuild_suite.bat
```diff
@@ -170,6 +170,14 @@ endlocal & set deleteIni=%deleteIni%
 
 if %deleteINI%==1 echo.[compiler list] >"%ini%"
 
+:ffmpegPath
+set "writePath=no"
+if %ffmpegPathINI%==0 (
+    set ffmpegPath="https://git.ffmpeg.org/ffmpeg.git"
+    ) else set ffmpegPath=%ffmpegPathINI%
+if %deleteINI%==1 set "writePath=yes"
+if %writePath%==yes echo.ffmpegPath=%ffmpegPath%>>%ini%
```
1480c1, line 179:
```suggestion
if %deleteINI%==1 echo.ffmpegPath=%ffmpegPath%>>%ini%
```
There doesn't seem to be an obvious reason to have the writePath variable here?

media-autobuild_suite.bat
```diff
@@ -170,6 +170,14 @@ endlocal & set deleteIni=%deleteIni%
 
 if %deleteINI%==1 echo.[compiler list] >"%ini%"
 
+:ffmpegPath
+set "writePath=no"
+if %ffmpegPathINI%==0 (
+    set ffmpegPath="https://git.ffmpeg.org/ffmpeg.git"
+    ) else set ffmpegPath=%ffmpegPathINI%
```
1480c1, line 177:
```suggestion
) else set ffmpegPath=%ffmpegPathINI%
```

media-autobuild_suite.bat
```diff
@@ -1740,6 +1748,17 @@ title MABSbat
...
+set nixdir=%nixdir::=%
+
+echo.ffmpegPath: %ffmpegPath%
+
+if [%ffmpegPath:~0,2%]==[..] (
+	set ffmpegPath=/%nixdir%/%ffmpegPath%
+	)
+
+echo.ffmpegPath: %ffmpegPath%
+
```
1480c1, line 1761: This whole logic might be better off being grouped in line 173-180 with a message of `Using <path> for FFmpeg` or something, rather than all the way down here 

Also, seems the two echos are redundant? Perhaps one was a debug print?

media-autobuild_suite.bat
```diff
@@ -1740,6 +1748,17 @@ title MABSbat
...
 if exist %build%\fail_comp del %build%\compilation_failed
 
+set nixdir=%instdir:\=/%
+set nixdir=%nixdir::=%
+
+echo.ffmpegPath: %ffmpegPath%
+
+if [%ffmpegPath:~0,2%]==[..] (
+	set ffmpegPath=/%nixdir%/%ffmpegPath%
+	)
```
1480c1, line 1758:
```suggestion
if [%ffmpegPath:~0,2%]==[..] (
    set ffmpegPath=/%nixdir%/%ffmpegPath%
)
```

media-autobuild_suite.bat
```diff
@@ -1751,8 +1770,8 @@ set compileArgs=--cpuCount=%cpuCount% --build32=%build32% --build64=%build64% ^
 --vvc=%vvc% --uvg266=%uvg266% --vvenc=%vvenc% --vvdec=%vvdec% --jq=%jq% --jo=%jo% --dssim=%dssim% ^
 --avs2=%avs2% --timeStamp=%timeStamp% --noMintty=%noMintty% --ccache=%ccache% --svthevc=%svthevc% ^
 --svtav1=%svtav1% --svtvp9=%svtvp9% --xvc=%xvc% --vlc=%vlc% --libavif=%libavif% --jpegxl=%jpegxl% ^
---autouploadlogs=%autouploadlogs%
-    set "noMintty=%noMintty%"
+ --ffmpegPath=%ffmpegPath% 
+    set "noMintty=%noMintty%"
```
1480c1, line 1774:
```suggestion
--autouploadlogs=%autouploadlogs% --ffmpegPath=%ffmpegPath% 
    set "noMintty=%noMintty%"
```

seems an accidental delete of another option, and an unnecessary trailing space

media-autobuild_suite.bat
1480c1: You may also want to echo out directions/infomation while this is being set (so in the `if %ffmpegPathINI%==0` block).

So something like 

you would need to add it inside the media-autobuild_suite.ini file like
```ini
ffmpegPath=https://github.com/softworkz/FFmpeg.git#branch=submit_qsv_sei
```
or something, else document this inside the README.md

I at least think the above would work, haven't checked out this branch and tested it yet

1480c1 (changes requested):
overall looks good

softworkz:
Maybe it makes sense to disable uploading of log files when a custom path is used, because you surely don't want to see such logs..?

1480c1:
> Maybe it makes sense to disable uploading of log files when a custom path is used, because you surely don't want to see such logs..?

Maybe, depends really on what failed, if something else other than FFmpeg failed, then it would still be good to know, although using a custom FFmpeg would exclude an issue from receiving support unless I feel like it.

But if you wish to go that route, then you may want to issue a warning that a `custom ffmpeg path will not be supported and autouploading of logs will be disabled` and set `autouploadlogsINI` to 2 if `ffmpegPathINI` != `defaultFFmpegPath` or something

softworkz:
Thanks for the suggestions. I will look into these in detail.

I made this patch years ago and continuously rebased. What I've submitted is working, but I've never put more thought into it since then :-)

softworkz:
> Maybe, depends really on what failed, if something else other than FFmpeg failed, then it would still be good to know, although using a custom FFmpeg would exclude an issue from receiving support unless I feel like it.

I can assure you that 99% of the compilation issues I had were none that you wanted to have seen :-) 

> But if you wish to go that route, then you may want to issue a warning that a `custom ffmpeg path will not be supported and autouploading of logs will be disabled` and set `autouploadlogsINI` to 2 if `ffmpegPathINI` != `defaultFFmpegPath` or something

Let me get the core part straight according to your suggestions, then you can think about which way you prefer, okay?

1480c1:
Sounds good to me

media-autobuild_suite.bat
softworkz: I believe that in a very early version I had tried to make this part of the interactive questioning procedure but later got away from doing that.

media-autobuild_suite.bat
softworkz: > This whole logic might be better off being grouped in line 173-180

For some obscure reason, the same commands don't work when placed there. 
I remembered this after I had tried to move it up there. I don't remember whether I ever figured out the reason. It's this statement

```
if [%ffmpegPath:~0,2%]==[..] (
    set ffmpegPath=/%nixdir%/%ffmpegPath%
)
```
which is always producing

`The syntax of the command is incorrect.`

But not when placed down in the range of 1700 .

softworkz:
I have updated the PR according to your suggestions. Only the moving of the lower statements didn't work (see my comment above).

media-autobuild_suite.bat
1480c1: I know that batch has a lot of horrible quirks with variables, disqualifying it from being a sane scripting language, let me see if there's a way, and I'll get back to this

softworkz: Maybe it's something about setlocal and delayed expansion.  
At least we know now how the two echo statements came to be there :laughing:

1480c1 (commented):
You may also want to consider adding a `if exist %ffmpegPathINI% set ffmpegPath=%ffmpegPathINI%` check or something similar when attempting to handle local paths

media-autobuild_suite.bat
```diff
@@ -170,6 +170,36 @@ endlocal & set deleteIni=%deleteIni%
...
+    echo -------------------------------------------------------------------------------
+    echo -------------------------------------------------------------------------------
+) else (
+    set ffmpegPath=%ffmpegPathINI%
+    echo -------------------------------------------------------------------------------
+    echo.
+    echo. Using ffmpeg path: %ffmpegPathINI%
+    echo.
+    echo -------------------------------------------------------------------------------
+)
```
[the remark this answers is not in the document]
softworkz: Done

media-autobuild_suite.bat
```diff
@@ -170,6 +170,36 @@ endlocal & set deleteIni=%deleteIni%
 
 if %deleteINI%==1 echo.[compiler list] >"%ini%"
 
+:ffmpegPath
+if %ffmpegPathINI%==0 (
+    set ffmpegPath="https://git.ffmpeg.org/ffmpeg.git"
```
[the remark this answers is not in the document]
softworkz: I thought that quotes would get eliminated but that was another script "language" I guess..

Done

softworkz:
Updated

media-autobuild_suite.bat
1480c1: I came up with https://github.com/1480c1/media-autobuild_suite/commit/4ce390ce6fec0d531d63b9310054653b11268238

and it seems to work locally,

first run after switching to the branch, reusing an old build
```bat
.\media-autobuild_suite.bat

 Using default ffmpeg source path: https://git.ffmpeg.org/ffmpeg.git

 If you want to use a custom source repository, add a line like this
 to media-autobuild_suite.ini:

     ffmpegPath=https://github.com/username/FFmpeg.git#branch=branchname

 or for a local repository like:

     ffmpegPath=../myrepos/ffmpeg

update autobuild suite
```

second run
```bat
.\media-autobuild_suite.bat
update autobuild suite
```

after modifying .\build\media-autobuild_suite.ini with `ffmpegPath=C:\Users\cddeg\FFmpeg`
```bat
.\media-autobuild_suite.bat

 Using ffmpeg path: C:\Users\cddeg\FFmpeg

update autobuild suite
```

after using `ffmpegPath=../FFmpeg
```
.\media-autobuild_suite.bat

 Using ffmpeg path: /D/FFmpeg

update autobuild suite
```

1480c1: That was made before your update, so you will need to adapt it as you see fit

softworkz: Besides the subroutine call, my update is identical.

For this...

```
set "resolvePath=%~dpnx1"
```

...I don't even want to know what it is and why it works.  :-)

1480c1: https://stackoverflow.com/a/18642270
> expands to the drive, path, name and extension of the first argument

softworkz: Will the result be relative to %instdir% when the script is called from a different dir?

1480c1: It will always be relative to instdir because of the cd at the top of the script

softworkz: Perfect!

softworkz: Updated with your changes included

media-autobuild_suite.bat
```diff
@@ -1930,3 +1973,7 @@ if not exist %instdir%\msys64\mingw%1\bin\gcc.exe (
 )
 endlocal
 goto :EOF
+
+:resolvePath
+set "resolvePath=%~dpnx1"
+goto :EOF
```
1480c1, line 1979: please preserve the newline at EOF

1480c1 (commented):
past that, lgtm

media-autobuild_suite.bat
softworkz: Damn I wondered whether there is supposed to be one, it just didn't look like there was any on GitHub:

softworkz: Done

1480c1:
LGTM, thanks

softworkz:
Cool! That means 1 down from my customizations...

Thanks a lot for your kind assistance!

Issue #2340, opened 2022-12-08
Proposal for Splitting out Reference Links


softworkz:
Hi,

I want to submit a PR that would make the management of dependency libraries easier and more transparent, but before doing so, I wanted to check back whether this would be agreeable in one or another way.

A problem with the current scripting is that the references to library sources are buried inside and spread all over the script with its 3k lines. It's tedious to follow changes and/or make changes or compare the states from two different points in time.  
Customization of library versions has often be requested in the past, but I understand that this wouldn't end well and would be a nightmare to maintain and support - so I don't want to go as far as that, but just make it a bit easier for those who want to do it (like me) at their own risk.

The change I'm proposing is really simple:

#### Replace all hard-coded source URLs with variables

Like this:

![image](https://user-images.githubusercontent.com/4985349/206343159-a22adf6c-5e25-4afe-a3c4-6bd27f3340d8.png)

#### And put the variable initialization...

![image](https://user-images.githubusercontent.com/4985349/206343414-a7b7c6e8-a998-41c1-af74-0e65133f7d6a.png)

...into a separate file - e.g. `dependency-links.sh`

Then, `media-suite_compile.sh` just needs to run it on start to get all variables set.  
The benefit would be to have all dependency links in one place and easy to review, compare, adjust, etc.

Would this be an acceptable PR? (or something similar in that direction)

Thanks,  
sw

1480c1:
personally okay with this

softworkz:
Cool, I'll prepare a PR.

A next step could be to think about whether patches should go there as well, maybe?

1480c1:
Unsure about the patches, unless it's something like a comma separated list of urls+git am/no gitam thing, as patches may often be temporary or semi-permanent depending on the content

softworkz:
> Unsure about the patches, unless it's something like a comma separated list of urls+git am/no gitam thing, as patches may often be temporary or semi-permanent depending on the content

I'm unsure myself. Probably time will tell whether there's a need to. What's on the good side is that the builds don't stop in case a patch is not applicable, so it's not a big problem when there's a version specified to which a patch doesn't apply.

I've submitted the PR now: #2341

PR #2341, opened 2022-12-12
Move repo links into separate file

softworkz:
Implements https://github.com/m-ab-s/media-autobuild_suite/issues/2340

This PR splits out all hard-coded links to source repositories into a separate file `media-suite_deps.sh` for better overview and possible customization (e.g. pin to a specific release instead of always pulling HEAD).

For a more detailed explanation please see here: [Proposal for Splitting out Reference Links](https://github.com/m-ab-s/media-autobuild_suite/issues/2340)

I have skipped the following for now:

- wget fetches
  These are mostly pulling old code which doesn't get updated anymore and/or requires a lot of patching
- Links related to building VLC
  Seems to be very experimental at this time
- Patches
  Generally, it might make some sense to also specifiy patches in the _deps file, but this would require more backend changes for handling (looping over 0, 1 or more patches to apply) than it would appear to be reasonable for including in this PR, and probably a bit more consideration than for this change alone.


softworkz:
To make sure that I haven't made any mistakes, I have created a little script/app for checking. It's C# for the sole reason that it takes me the least time:

```c#
        public static void Main(string[] args)
        {
            if (args.Length != 3)
            {
                Console.WriteLine("Usage: variablesfile inputfile outputfile");
            }

            var dic = File.ReadLines(args[0])
                .Select(e => e.Split('=', 2))
                .Where(e => e.Length > 1)
                .ToDictionary(e => e[0], e => e[1]);

            var source = File.ReadAllText(args[1]);

            foreach (var (key, value) in dic)
            {
                var find = "\"$" + key + '\"';
                if (!source.Contains(find))
                {
                    Console.WriteLine("Error: Could not find " + find);
                }

                source = source.Replace(find, '\"' + value + '\"');
            }

            File.WriteAllText(args[2], source);
        }
```

It simply reads all variables from the the `media-suite_deps.sh` file and uses them for replacement in `media-suite_compile.sh`:

```shell
ReplaceStrings.exe media-suite_deps.sh media-suite_compile.sh out.sh
```

Finally, I compared `out.sh` with the original `media-suite_compile.sh` to make sure it's the same.

build/media-suite_deps.sh
```diff
@@ -0,0 +1,107 @@
...
+SOURCE_REPO_SPIRV_CROSS=https://github.com/KhronosGroup/SPIRV-Cross.git
+SOURCE_REPO_SPIRV_TOOLS=https://github.com/KhronosGroup/SPIRV-Tools.git
+SOURCE_REPO_SPIRV_HEADERS=https://github.com/KhronosGroup/SPIRV-Headers.git
+SOURCE_REPO_MPV=https://github.com/mpv-player/mpv.git
+SOURCE_REPO_LIBMXF=https://gitlab.com/m-ab-s/libmxf.git
+SOURCE_REPO_LIBMXFPP=https://gitlab.com/m-ab-s/libmxfpp.git
+SOURCE_REPO_LIBBMX=https://gitlab.com/m-ab-s/bmx.git
+SOURCE_REPO_NEON=https://github.com/notroj/neon.git
+SOURCE_REPO_LIBMUSICBRAINZ=https://github.com/wiiaboo/libmusicbrainz.git
+SOURCE_REPO_CYANRIP=https://github.com/cyanreg/cyanrip.git
```
1480c1, line 107: perhaps it may be better to sort these alphabetically to be able to quickly look for a repo, so `sort -uo media-suite_deps.sh{,}`

softworkz: I was unsure. The list was "in order of appearance".

But it's sorted now.

softworkz:
I also wasn't sure when to use LIB and when not. I made the decision based on the name of the PKGCONF (\*.pc) file (where available).

softworkz:
Meanwhile I've also run a "from-scratch" compilation with this PR applied which completed successfully:

(64bit with almost everything enabled excepting VLC and excepting the "not recommended" parts)

build/media-suite_deps.sh
```diff
@@ -0,0 +1,107 @@
...
+SOURCE_REPO_VPX=https://chromium.googlesource.com/webm/libvpx
+SOURCE_REPO_VULKANLOADER=https://github.com/KhronosGroup/Vulkan-Loader.git
+SOURCE_REPO_VVC=https://gitlab.com/m-ab-s/VVCSoftware_VTM.git
+SOURCE_REPO_X264=https://code.videolan.org/videolan/x264.git
+SOURCE_REPO_X265=https://bitbucket.org/multicoreware/x265_git.git
+SOURCE_REPO_XAVS2=https://github.com/pkuvcl/xavs2.git
+SOURCE_REPO_XAVS=https://github.com/Distrotech/xavs.git
+SOURCE_REPO_XVC=https://github.com/divideon/xvc.git
+SOURCE_REPO_XVID=https://github.com/m-ab-s/xvid.git
+SOURCE_REPO_ZIMG=https://github.com/sekrit-twc/zimg.git
```
1480c1, line 107: newline missing

1480c1:
> I also wasn't sure when to use LIB prefix and when not. I made the decision based on the name of the PKGCONF (*.pc) file (where available).

good idea

build/media-suite_deps.sh
Biswa96: editorconfig or clang-format mey help in future.

softworkz: > newline missing

I could have sworn that I already noticed and pushed an update to fix it. Must have dreamt it..
Fixed.

softworkz: > **editorconfig** or clang-format mey help in future.

Do you have one I could use?

softworkz: > > newline missing
> 
> I could have sworn that I already noticed and pushed an update to fix it. Must have dreamt it.. Fixed.

Ah - mystery solved. I actually did fix it (https://github.com/m-ab-s/media-autobuild_suite/compare/96aaf5ff83781b9c85e662a842c0e5dde5fe5dae..6d1dd135fdda14f6304543ede6a0e9390ee2066b) but then it got lost again when I had done the sorting.

1480c1:
Thanks

Issue #2358, opened 2023-01-05
Enable d3d11 by disabling the SIRV meson patch


RivenSkaye:
``` bash
Paste output of the suite here
```

[logs.zip](https://github.com/jb-alvarado/media-autobuild_suite)

Issue #2346, opened 2022-12-20
mpv build stalls after config/build/installation


rahqbo:
The mpv build stalls after each step and only continues with ctl+c. The log contains a build run where I ctl+c'd after config and then during build to create the log.

When finished in this configuration, the mpv.exe is about twice the size it normally is.

[logs.zip](https://github.com/m-ab-s/media-autobuild_suite/files/10268034/logs.zip)

rahqbo:
Update: mpv.exe is normal sized, but the ctl+c issue persists.

basilisk0315:
I have the same here.

snowbie:
Same. Reinstalling didn't work and letting it run for a few hours didn't make any differences.

1480c1:
Looking at it locally, it seems that even though it finished configuring, waf is curiously not exiting

RivenSkaye:
Hitting the same issue here, I'll try sending a SIGQUIT orSIGKILL, but would it be possible to rely on the Meson toolchain for at least until this is fixed?
And is there any sight on what causes WAF to hang up?

1480c1:
None personally, have no experience with waf to know where to begin, planning on looking into moving to meson either way. Main concern right now would be command line difference messing with the mpv_options.txt, might need to detect waf option and move the old one and request the user to remake the new one with any old options they may have had

RivenSkaye:
Luckily I've been through the mpv through Meson rodeo before o Arch, I'll see if I can find the differences in the options they expose and if it's possible to reuse existing configs. Logically the command line API should be the only real difference and with a bit of luck we can reuse parts of the AUR package `mpv-git`. Here's to hoping we can fix it before next year hits 🎉

basilisk0315:
Sorry @RivenSkaye too late. It's 2023 here already!

Greetings from Australia!

RivenSkaye:
Considering I've got about 20 mins left and I'm not yet any closer to fixing it, we'll have to shoot for migrating before 2024 instead. Much more feasible, happy new year folks!

32-bit-link:
Greeting from 2023!
Sadly experiencing the same issue here as well.

Biswa96:
Can anyone check if excluding the source & build directory in Windows Defender or in other antivirus software helps?

RivenSkaye:
Excluding it doesn't change anything for me

Ajaja:
Killing stalled python.exe in the task manager helps.

[...]

PR #2359, opened 2023-01-05
Use Meson for mpv instead of Waf

RivenSkaye:
Pull request that translates Waf options to their Meson equivalents at runtime, fixing #2346
Also ensures any flags guaranteed to be illegal on Windows are disabled, with a warning sent to the user.

Still has one call to `waf distclean`, since that doesn't hang up and will allow for users to keep using existing envs. That call can be removed later, when it can be reasonably assumed nobody who's remotely up-to-date has Waf artifacts left that might interfere with the build.
A side effect of this change is the removal of the option to build `mruby` support into mpv, as the relevant branch has been dead since 2017 and can't be built with Meson.
Work on this also brought to light something else, due to Meson being a lot less forgiving when it comes to impossible combinations and configure flags. There is currently no way to explicitly enable `d3d11` and `spirv-cross` flags due to #2358 blocking the build of `spirv-cross-c-shared`. It seems like all of the Vulkan stuff that relies on SPIR-V works so long as the shared tools are built, which still happens with the spirv-meson patch.

I'm opening this PR as a **draft** because this still needs some considerable cleanup. Things like the extra `LDFLAGS` being set in order to allow linking mpv to libarchive without undefined symbol references. I'm not exactly sure why these changes were needed to make mpv build properly, as I got it to work just fine standalone. This only seemed to be a problem within the context of the Suite.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2384,7 +2384,6 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
         hide_conflicting_libs
         create_ab_pkgconfig
 
-        log bootstrap /usr/bin/python bootstrap.py
         if [[ -d build ]]; then
             # Temporarily keep this until Waf is reasonably phased out.
             # Not all users will run this weekly, so keep the cleanup for now.
@@ -2447,15 +2446,20 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
         # The alternative is having users only specify everything after it in the config, e.g.
         # `lua=luajit` and `cdda=enabled` (optionally omitting `=enabled` for on/off features)
         local meson_opts=()
+        local default_lib="both"
         local option
         for option in "${MPV_OPTS[@]}"; do
-            do_simple_print "Processing $option"
             # Process Waf flags into Meson options.
             # starting with boolean flags
-            if [[ $option =~ ^--enable-(gpl|cplayer|libmpv-shared|build-date|tests|ta-leak-report) ]]; then
+            if [[ $option =~ ^--enable-(gpl|cplayer|build-date|tests|ta-leak-report) ]]; then
                 meson_opts+=("${option/--enable-/-D}=true")
-            elif [[ $option =~ ^--disable-(gpl|cplayer|libmpv(-shared)?|build-date|tests|ta-leak-report) ]]; then
-                meson_opts+=("${${option/--disable-/-D}/libmpv-shared/libmpv}=false")
+            elif [[ $option =~ ^--disable-(gpl|cplayer|build-date|tests|ta-leak-report) ]]; then
+                meson_opts+=("${option/--disable-/-D}=false")
+            elif [[ $option =~ ^--enable-libmpv(-shared)? ]]; then
+                meson_opts+=("-Dlibmpv=true")
+            elif [[ $option =~ ^--disable-libmpv(-shared)? ]]; then
+                meson_opts+=("-Dlibmpv=false")
+                default_lib="static"
             # features in the misc section
             elif [[ $option =~ ^--enable-(cdda|cplugins|dvbin|dvdnav|iconv|javascript|lcms2|libarchive|libavdevice|libbluray|pthread-debug|rubberband|sdl2|sdl2-gamepad|stdatomic|uchardet|uwp|vapoursynth|vector|win32-internal-pthreads|zimg|zlib) ]]; then
                 meson_opts+=("${option/--enable-/-D}=enabled")
@@ -2509,7 +2513,7 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
                 # Remove the `-Dlua=enabled` flag if present, since the order of flags matters.
                 # - `-Dlua=lua5.2 -Dlua=enabled` makes Meson behave like only `-Dlua=enabled` was specified;
                 # - `-Dlua=enabled -Dlua=lua5.2` makes Meson take lua5.2 if available, erroring otherwise;
-                meson_opts=("${meson_opts[@]/-Dlua=enabled}")
+                meson_opts=("${meson_opts[@]/ -Dlua=enabled}")
             elif [[ $option =~ ^--enable-lua ]] && ! [[ "${meson_opts[@]}" =~ -Dlua= ]]; then
                 meson_opts+=("-Dlua=enabled")
             fi
@@ -2521,19 +2525,18 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
         CFLAGS+=" ${mpv_cflags[*]} -Wno-int-conversion" LDFLAGS+=" ${mpv_ldflags[*]}" \
             RST2MAN="${MINGW_PREFIX}/bin/rst2man" \
             RST2HTML="${MINGW_PREFIX}/bin/rst2html" \
-            RST2PDF="${MINGW_PREFIX}/bin/rst2pdf2" \
-            PKG_CONFIG="pkgconf --keep-system-libs --keep-system-cflags $LOCALDESTDIR/bin/ab-pkg-config" CC=${CC/ccache /}.bat CXX=${CXX/ccache /}.bat
+            RST2PDF="${MINGW_PREFIX}/bin/rst2pdf2"
         mkdir build
-        log "meson" meson setup build
-        log "meson" meson configure --default-library=static --buildtype=release --prefix="$LOCALDESTDIR" --backend=ninja --bindir=bin-video "${meson_opts[@]}"
+        PKG_CONFIG="pkgconf --keep-system-libs --keep-system-cflags" CC=${CC/ccache /}.bat CXX=${CXX/ccache /}.bat \
+            log "meson.setup" meson setup build "${meson_opts[@]}" --prefix="${LOCALDESTDIR}" --bindir="${LOCALDESTDIR}/bin-video" --default-library="${default_lib}" --buildtype=release
         extra_script post configure
 
-        extra_script pre ninja
-        log "build" ninja
-        extra_script post ninja
+        extra_script pre build
+        log "meson.compile" meson compile -C build
+        extra_script post build
 
         extra_script pre install
-        cpuCount=1 log "install" ninja install
+        cpuCount=1 log "meson.install" meson install -C build
         extra_script post install
 
         if ! files_exist libavutil.a; then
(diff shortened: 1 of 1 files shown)


build/media-suite_compile.sh
```diff
@@ -2432,27 +2438,107 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
...
+                do_simple_print "${orange}Disabling option '${option}' as it's not usable on Windows"'!'"${reset}"
+            elif [[ $option =~ ^--enable-(jack|openal|opensles|sdl2-audio|wasapi) ]]; then
+                meson_opts+=("${option/--enable-/-D}=enabled")
+            elif [[ $option =~ ^--disable-(jack|openal|opensles|sdl2-audio|wasapi) ]]; then
+                meson_opts+=("${option/--disable-/-D}=disabled")
+            # Video corner, same as with audio
+            elif [[ $option =~ ^--enable-(cocoa|drm|egl-android|*-drm|*-wayland|*-x11|*-cocoa|gbm|rpi|vdpau|vdpau-*|vaapi|vaapi-*|wayland|x11|xv) ]]; then
+                meson_opts+=("${option/--enable-/-D}=disabled")
+                do_simple_print "${orange}Disabling option '${option}' as it's not usable on Windows"'!'"${reset}"
+            elif [[ $option =~ ^--enable-(caca|d3d11|direct3d|egl|egl-*|plain-gl|gl|gl-*|jpeg|libplacebo|sdl2-video|shaderc|sixel|spirv-cross|vulkan) ]]; then
```
RivenSkaye, line 2480: I'll have to change this one a bit, as the suite doesn't include sixel (yet) and there's no release that can be grabbed from the repos

build/media-suite_compile.sh
```diff
@@ -2432,27 +2438,107 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
...
+            # Mac-only options, we might as well catch and squash them
+            elif [[ $option =~ ^--enable-(macos-*|swift-*) ]]; then
+                meson_opts+=("${option/--enable-/-D}=disabled")
+                do_simple_print "${orange}Disabling option '${option}' as it's not usable on Windows"'!'"${reset}"
+            # Manpages corner
+            elif [[ $option =~ ^--enable-(html-build|manpage-build) ]]; then
+                meson_opts+=("${option/--enable-/-D}=enabled")
+            elif [[ $option =~ ^--disable-(html-build|manpage-build) ]]; then
+                meson_opts+=("${option/--disable-/-D}=disabled")
+            elif [[ $option =~ ^--enable-pdf-build ]]; then
```
RivenSkaye, line 2501: Remnant of the old code, which silently and unconditionally disabled PDF doc builds, but this needs to be tested

build/media-suite_compile.sh
```diff
@@ -2432,27 +2438,107 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
...
             RST2HTML="${MINGW_PREFIX}/bin/rst2html" \
-            RST2PDF="${MINGW_PREFIX}/bin/rst2pdf2" \
-            PKG_CONFIG="$LOCALDESTDIR/bin/ab-pkg-config" \
-            log configure /usr/bin/python waf configure \
-            "--prefix=$LOCALDESTDIR" "--bindir=$LOCALDESTDIR/bin-video" \
-            "${MPV_OPTS[@]}"
+            RST2PDF="${MINGW_PREFIX}/bin/rst2pdf2"
+        mkdir build
+        PKG_CONFIG="pkgconf --keep-system-libs --keep-system-cflags" CC=${CC/ccache /}.bat CXX=${CXX/ccache /}.bat \
+            log "meson.setup" meson setup build "${meson_opts[@]}" --prefix="${LOCALDESTDIR}" --bindir="${LOCALDESTDIR}/bin-video" --default-library="${default_lib}" --buildtype=release
```
RivenSkaye, line 2532: log names are non-standard atm, but I expect there to be more changes required to the Meson calls.
Kept this order for now so that it
1) doesn't break existing `mpv_extra.sh` setups
2) I wanted it to work first, and then adapt it to use the helper functions later

build/media-suite_compile.sh
```diff
@@ -2432,27 +2438,107 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
...
         extra_script pre build
-        log build /usr/bin/python waf -j "${cpuCount:-1}"
+        log "meson.compile" meson compile -C build
         extra_script post build
 
         extra_script pre install
-        log install /usr/bin/python waf -j1 install ||
-            log install /usr/bin/python waf -j1 install
+        cpuCount=1 log "meson.install" meson install -C build &&
+        mv "${LOCALDESTDIR}/bin-video/libmpv-2.dll" "$LOCALDESTDIR/bin-video/mpv-2.dll"
```
RivenSkaye, line 2541: I can't find a way to tell Meson I don't want the output file to be named `libmpv-2.dll`, hence the manual move

build/media-suite_compile.sh
```diff
@@ -2464,7 +2550,10 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
             done
         fi
 
-        unset mpv_ldflags replace PKGCONF_STATIC
+        LDFLAGS="$old_ldf"
+        replace=" ${mpv_cflags[*]} -Wno-int-conversion"
+        CFLAGS="${CFLAGS/$replace}"
+        unset mpv_ldflags replace PKGCONF_STATIC meson_opts default_lib old_ldf
```
RivenSkaye, line 2556: I'm not entirely happy with this cleanup, but it works:tm:
I'll look into `extra-cflags` options and similar for Meson

build/media-suite_compile.sh
```diff
@@ -2432,27 +2438,107 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
...
+        # since users editing a config may easily slip up with `-D` prepended to every line.
+        # The alternative is having users only specify everything after it in the config, e.g.
+        # `lua=luajit` and `cdda=enabled` (optionally omitting `=enabled` for on/off features)
+        local meson_opts=()
+        local default_lib="both"
+        local option
+        for option in "${MPV_OPTS[@]}"; do
+            # Process Waf flags into Meson options.
+            # starting with boolean flags
+            if [[ $option =~ ^--enable-(gpl|cplayer|build-date|tests|ta-leak-report) ]]; then
```
RivenSkaye, line 2454: TA Leak can only be turned off by also adjusting an env var; perhaps make some accessible entry point for people to inject the option from a custom patch instead of setting it here?

build/media-suite_compile.sh
RivenSkaye, line 2480: There's also some things to note about SPIRV-Cross, see issue linked in the initial comment on the PR

RivenSkaye (commented):
Partial feedback for myself, some questions for the maintainers.
I turned on maintainer edits for the PR, so if there's things you'd like to tackle, feel free to add it to the PR

Quackdoc:
https://github.com/mpv-player/mpv/pull/11143

vt-idiot:
Does this work?

RivenSkaye:
I mean, last I touched it I managed to produce a build. If that counts as working, then yes. But for all intwnts and purposes it needs a fair amount of TLC and I currently just cannot spare the time to wrangle with upstream changes on mpv

vt-idiot:
I did get past the OpenAL issues by disabling it in ffmpeg, rebuilding, and going again. libvulkan's DllMain is going to be an issue for building libmpv-2.dll unless there's a different way of patching vulkan-loader. I put a couple hours into it and couldn't make heads or tails of it.

RivenSkaye:
Huh, last time I ran a build off upstream I couldn't get vulkan to work properly either, and it has also been a bit iffy on Arch to get everything to play nice.

And yes, spending a few hours on it sounds like pretty much every step of the way I already had into this PR. Hence why time is an issue, I'd almost miss my student years with no household of my own to run...

vt-idiot:
Might be worth looking at https://github.com/shinchiro/mpv-winbuild-cmake/tree/master/packages if you ever get the chance.

1480c1:
I'm thinking perhaps we just go the nuke route and just move the old option file and make a new one.

The main thing I'm concerned about is the whole enable/disable and using that logic with stuff that might not adhere to it with meson, like instead of enable/disable, some options might be bools. We may instead need to introduce a different set of stuff for meson and like `mpv_enable()` and `mpv_true()` or something with different arrays for the different types.

vt-idiot:
> I'm thinking perhaps we just go the nuke route and just move the old option file and make a new one.

Not a bad idea. One issue (maybe 2?) I did run into is that even in a fresh environment, libvulkan causes it to fail because of the DllMain definition.

Also the way spirv-cross currently builds probably still needs to be addressed, I'm not sure why it still uses all that meson patchwork when there is now a build system in-place upstream.

1480c1:
realized there may be more pains with meson considering the user could do something like
```bash
-D option=enable
```

Need to do a lot more worrying about parsing on bash's side, past just slurping up the arguments

RivenSkaye:
That's why my current rough draft is just parsing the existing options files. I say we butcher the config generation in three places.

1. Initial setup, where we only write non-default features (and explicit non-options due to Windows constraints)
2. Before we finalize setup and start the first build, based on relevant environment variables and args passed with a `-D` prefix
3. Before the actual build of mpv itself, once all patches and additional user scripts have ran and set their info. If this changes settings in a breaking way, we ask the user to check their custom scripts.

If we don't yet, it might be worthwhile to set a flag to see if the user added any custom scripts (perhaps ``echo `cygpath -wa $script_path` >> .userscripts.log`` or something?)

We'd then probably end up with a glorified commandline in a file, but we at least wouldn't have to worry about parsing the contents anymore

RivenSkaye:
Burned a few hours trying to merge upstream changes into my fork, but the amount of work needed to get it merged is a nightmare and all it takes is a small mistake to leave the script in a horribly broken state.  
At this point I'm wondering if it would be better to clean my fork and redo my changes, taking into account other changes made since I opened the PR, or if I should somehow piece together the merge so I can move forward on this PR.

1480c1:
Finally got around to working on mpv and meson due to a bunch of breaking changes from FFmpeg's side https://github.com/m-ab-s/media-autobuild_suite/commit/96005f0e215108ab08d0be1f78ee2d1cca58aac9

RivenSkaye:
Oh that's absolutely amazing. I was still trying to get upstream changes merged just to get to a workable state to continue this. Guess I don't have to anymore then.  
Is there still a need to parse the old options file for backward compat/seamless updating?

1480c1:
I've given up on doing some kind of waf -> meson argument conversions due to the complexity and lack of time. Hopefully, if I got the batch stuff correct, it should rename the old options and the user can adjust the new one as they want. If you feel like it's worth investing time in a script or something to do that conversion, feel free to do so.

PR #2369, opened 2023-02-03
Early Exit: Provide a way to cleanly exit the build at defined points

softworkz:
The build process can take several hours. In case of build errors, one is usually tempted to fix the error and re-run the build, but continuing on failed state with changes does not exactly guarantee the same result (or even a successful result) as when re-builing from scratch with those changes applied.

Having exit points like added by this commit allows to perform a build incrementally and cleanly from point-to-point, when the intermediate results are saved each time.

This is also very useful for setting up automated builds via GH actions or DevOps where intermediate results are cached only when no errors have occurred.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -73,6 +73,7 @@ while true; do
     --svtvp9=* ) svtvp9=${1#*=} && shift ;;
     --xvc=* ) xvc=${1#*=} && shift ;;
     --vlc=* ) vlc=${1#*=} && shift ;;
+    --exitearly=* ) exitearly=${1#*=} && shift ;;
     # --autouploadlogs=* ) autouploadlogs=${1#*=} && shift ;;
     -- ) shift && break ;;
     -* ) echo "Error, unknown option: '$1'." && exit 1 ;;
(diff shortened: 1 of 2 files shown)


build/media-suite_compile.sh
```diff
@@ -655,6 +666,11 @@ if [[ $ffmpeg != no ]] && enabled libzimg &&
     do_checkIfExist
 fi
 
+if [[ $exitearly = EE3 ]]; then
+    do_simple_print -p '\n\t'"${orange}Exit due to env var MABS_EXIT_EARLY set to EE3"
+    return
+fi
+
 set_title "compiling audio tools"
 do_simple_print -p '\n\t'"${orange}Starting $bits compilation of audio tools${reset}"
```
1480c1, line 675: I wonder if it would be better to just split the `build()` function into multiple sub-functions for each section and keep the `build()` function as a wrapper and consolidate the early exit checks into that wrapper instead

1480c1:
Going to go ahead and merge this, I'll think about the splitting up later

softworkz:
Thanks!

GyanD:
Docs missing.

PR #2529, opened 2023-10-04
fix libavif,libbluray,libbrotli and add a tad more security

hydra3333:
To get it to build static x64 only (without vulkan) I had to do a few patches per this pull (libavif,libbluray,brotli) :
...
and then disable vulkan and libplacebo in ffmpeg_options.txt.

I couldnt find a way to disable vulkan/libplacebo in the initial questions/answers first run, so I started it and let it go down to the point where it told me I could edit ffmpeg_options.txt whereupon I did and changed these:
```
#--enable-vulkan 
#--enable-libplacebo
#--enable-libglslang
--enable-libopenh264
```
to
```
--disable-vulkan
--disable-libplacebo
--disable-libglslang
--disable-libopenh264
```
(they may not have been commented out, if they were I removed the leading #)
If I saw it right, some things had a bit of a dependency on vulkan so they had to be disabled.
The --disable-libopenh264 was added later because I guess I did something wrong - when I ran ffmpeg it would not run and kept prompting me saying it could not find libopenh264.dll :(

No doubt the author and other good people will do much better fixes and whatnot, these are just a few hacks which worked for me for now for a static x64 build.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1958,7 +1958,7 @@ if { { [[ $ffmpeg != no ]] && enabled_any vulkan libplacebo; } ||
     do_patch "$_mabs/0001-pc-remove-CMAKE_CXX_IMPLICIT_LINK_LIBRARIES.patch" am
     do_patch "$_mabs/0002-loader-loader_windows-Static-linking-hacks.patch" am
     do_patch "$_mabs/0003-loader-CMake-related-static-hacks.patch" am
-    do_patch "$_mabs/0004-loader-Add-private-libs-to-pc-file.patch" am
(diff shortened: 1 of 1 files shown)


1480c1:
I will pull in the changes relating to fstack-protector as I'm already working on changing the env flags. However, I will wait on the other commits until I can verify on if they build or not, as I haven't gotten a compilation to get to those yet. Although I think libavif should have been fixed https://github.com/AOMediaCodec/libavif/issues/1647

hydra3333:
Cool.

There has to be a better way to fix libbluray as well. It must have been built here for a long time without the addition of the libraries, don't know why/how as other build system(s) had them added for a while ... something relevant must have changed here but I do not know what.

Issue #2580, opened 2024-01-19
mingw64 [vorbis] build vorbis fails


job-so:
Hi Folks,

using https://github.com/m-ab-s/media-autobuild_suite/commit/4b263b7915da5b21e6f3d32de22f555dc25255ec

Vorbis does not compile due to ogg missing. (no standalone binaries requested)
Had to manually preinstall ogg library : `pacman -S mingw-w64-x86_64-libogg`

Same Issue as #2491 ? 

``` 	
Warning: We will not accept any issues lacking any form of logs or logs.zip!
media-autobuild_suite  (64bit)	
	Starting 64bit compilation of all tools
Imported default FFmpeg options from .bat
License: nonfree
Imported default mpv options from .bat
media-autobuild_suite  (64bit): compiling global tools	
	Starting 64bit compilation of global tools
media-autobuild_suite  (64bit): compiling audio tools	
	Starting 64bit compilation of audio tools
  Running git update for vorbis...
┌ vorbis git  ......................................... [Recently updated]
├ Running autogen...
├ Running configure...
Likely error (tail of the failed operation logfile):
checking for cos in -lm... yes
checking for pthread_create in -lpthread... yes
checking pkg-config is at least version 0.9.0... yes
checking for OGG... no
checking for Ogg... no
*** Could not run Ogg test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occured. This usually means Ogg was incorrectly installed
*** or that you have moved Ogg since it was installed.
configure: error: must have Ogg installed!
configure failed. Check D:/ETALON/ffmpeg/media-autobuild_suite/build/vorbis-git/build-64bit/ab-suite.configure.log
This is required for other packages, so this script will exit.
  Creating diagnostics file...
```
[logs-20240119_1430.zip](https://github.com/m-ab-s/media-autobuild_suite/files/13991038/logs-20240119_1430.zip)

PR #2581, opened 2024-01-19
fix: #2580 remove standalone condition for ogg compilation

job-so:
The Ogg library is required to compile Vorbis.
Ogg is only compiled when standalone binaries were requested.

Vorbis compilation fails as ogg is missing when standalone librairies are not resqested

=> remove the standalone condition for compiling ogg

Corrections #2580


1480c1:
Thanks

PR #2648, opened 2024-04-18
Curl url

vt-idiot:
Moves curl URL into `media-suite_deps.sh` instead of hardcoding it into the build script.

This makes it easier to override with a specific tag/branch to prevent wasting an hour recompiling curl while troubleshooting other m-ab-s issues since there absolutely _had_ to be a second, separate commit to curl/master to fix documentation or something equally silly.


1480c1:
Thanks

PR #2668, opened 2024-05-05
cargo-c: install msys2 package instead of compiling from source

### LINKED ISSUE #2667: [mingw32|mingw64] cargo-c has to be compiled in every run due to "missing files"
Due to issues with my internet connection, I have to run the script several times to get it to properly compely. Th issue is that every time it reaches cargo-git (which it had also built already for at least two prior libraries) it says "missing files", redownloads it and successfully recompiles it EVERY SINGLE TIME.

[logs.zip](https://github.com/m-ab-s/media-autobuild_suite/files/15213156/logs.zip)

woot000:
Rust is installed via msys2's package now, so install cargo-c as a package too. /opt/cargo/bin is deleted on every update, so this prevents repetitive recompilation. 

Fixes #2667

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1118,14 +1118,7 @@ if { [[ $dav1d = y ]] || [[ $libavif = y ]] || { [[ $ffmpeg != no ]] && enabled 
     do_checkIfExist
 fi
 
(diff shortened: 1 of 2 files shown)


1480c1:
Thanks, I overlooked that as the potential reason for the issue

Issue #2692, opened 2024-05-21
Request: libavif - avifenc may support SVT-AV1 encoder in 64-bit build


LigH-de:
When building libavif, avifenc already gets AOM and rav1e enabled as available AV1 encoders if the suite compiles them. SVT-AV1 may also be supported; but it is only built in the Win64 branch, so adding it to available encoders in avifenc would also only be suitable for its 64-bit version.

gitoss:
It's nice to see all 3 av1 encoders included in libavif through m-a-s.

Is there a comparison of the resulting avif quality anywhere between aom, rav1e and svt anywhere?

Personally, rav1e seems quite competitive visually, but I didn'd do any synthetic benchmarks and didn't chec svt yet.

LigH-de:
The issues tracker is not suitable for discussions. You may like to ask in the doom9 forum instead.

PR #2693, opened 2024-05-21
Add SVT-AV1 codec to libavif

LigH-de:
SVT-AV1 encoder can be used in 64-bit build of avifenc

Fixes https://github.com/m-ab-s/media-autobuild_suite/issues/2692


1480c1:
Thanks

gitoss:
The compilation order is wrong - currently, svt-av1 is after libavif, i.e. on the first run svt-av1 isn't picked up by libavif, but needs another go.

Issue #2718, opened 2024-06-14
opus cant find libraries


cantrell216:
Likely error (tail of the failed operation logfile):
Making all in .
make[2]: Entering directory '/build/opus-tools-git/build-64bit'
  CCLD     opusenc.exe
C:/ffmpeg/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find @FLAC_STATIC_LIBS@: No such file or directory
collect2.exe: error: ld returned 1 exit status
make[2]: *** [Makefile:689: opusenc.exe] Error 1
make[2]: Leaving directory '/build/opus-tools-git/build-64bit'
make[1]: *** [Makefile:1112: all-recursive] Error 1
make[1]: Leaving directory '/build/opus-tools-git/build-64bit'
make: *** [Makefile:532: all] Error 2
make failed. Check C:/ffmpeg/build/opus-tools-git/build-64bit/ab-suite.make.log
This is required for other packages, so this script will exit.

CPPFLAGS: 
CFLAGS: -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 -mthreads
CXXFLAGS: -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1
LDFLAGS: -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 -static-libgcc -static-libstdc++
make -j6 
make  all-recursive
make[1]: Entering directory '/build/opus-tools-git/build-64bit'
Making all in .
make[2]: Entering directory '/build/opus-tools-git/build-64bit'
  CC       src/opusenc-opus_header.o
  CC       src/opusenc-opusenc.o
  CC       src/opusenc-tagcompare.o
  CC       src/opusenc-audio-in.o
  CC       src/opusenc-diag_range.o
  CC       src/opusenc-flac.o
  CC       win32/opusenc-unicode_support.o
  CC       src/opusdec-opus_header.o
  CC       src/opusdec-wav_io.o
  CC       src/opusdec-wave_out.o
  CC       src/opusdec-opusdec.o
  CC       src/opusdec-resample.o
[...]

PR #2720, opened 2024-06-16
FLAC autoconf to cmake migration and some minor change

L4cache:
Fixes: #2718


L4cache:
This seems to cause the later check (running the suite again) to get weird and install flac from pacman...
May need to further investigate

L4cache:
> This seems to cause the later check (running the suite again) to get weird and install flac from pacman... May need to further investigate

Actually... that's what the logic here does, with or without this migration to cmake.

PR #2768, opened 2024-09-05
Fix the clang pthread problem of liblc3 standalone tools and adjust some default ffmpeg options in bat

L4cache:

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1027,7 +1027,7 @@ if [[ $ffmpeg != no ]] && enabled liblc3 &&
     do_uninstall "${_check[@]}"
     if [[ $standalone = y ]]; then
         _check+=(bin-audio/{d,e}lc3.exe)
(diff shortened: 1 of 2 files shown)


1480c1:
@L4cache can you take a look at https://github.com/m-ab-s/media-autobuild_suite/commit/a294078174530a3d310422ec784c52ac5bb3f494#commitcomment-146281674 as well?

Seems to stem from https://github.com/intel/libvpl/blob/5f6bd8a1e753c8f63a3fd8b36894d6968b808a6d/cmake/InstallDirs.cmake#L11-L20 where it should just be using the `include(GNUInstallDirs)` line below for anything that is `NOT MSVC`. We can just probably just apply this patch from msys2 https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-libvpl/0003-cmake-fix-32bit-install.patch

1480c1:
Thanks

PR #2774, opened 2024-10-03
vulkan-loader: remove windows-Static-linking-hacks.patch

LigH-de:
Patch failed, built anyway. So it is probably implemented now.


1480c1:
Looking at what changed upstream, it indeed may need a more careful eye check to see if it's really necessary, and unfortunately I'm figuratively blind. I doubt it's actually implemented since upstream has said before that windows and static linking is fully unsupported and shouldn't be used, but I haven't checked in a while.

Issue #2516, opened 2023-09-24
[add] libheif


zedd-0:
hi, is it possible to add libheif to this script?
https://github.com/strukturag/libheif

hydra3333:
I don't recall if ffmpeg supports those formats already in some manner ... 
Ah. It seems ffmpeg itself does not support heif
https://stackoverflow.com/questions/65435536/convert-a-heif-file-to-png-jpg-using-ffmpeg

After quick google, I have not yet spotted anything in ffmpeg to say it supports libheif and a reasonable guess would be that ffmpeg needs to be updated to support it.

In fact https://github.com/strukturag/libheif#software-using-libheif does not list ffmpeg as software which uses libheif.

On that basis, since MABS is a _build system_ for ffmpeg and dependencies, I guess that no change to MABS would achieve what you are seeking ?

Lack of heif support seems to be a long overdue shortcoming in ffmpeg, based on TRAC comments.

L4cache:
> I don't recall if ffmpeg supports those formats already in some manner ... Ah. It seems ffmpeg itself does not support heif https://stackoverflow.com/questions/65435536/convert-a-heif-file-to-png-jpg-using-ffmpeg
> 
> After quick google, I have not yet spotted anything in ffmpeg to say it supports libheif and a reasonable guess would be that ffmpeg needs to be updated to support it.
> 
> In fact https://github.com/strukturag/libheif#software-using-libheif does not list ffmpeg as software which uses libheif.
> 
> On that basis, since MABS is a _build system_ for ffmpeg and dependencies, I guess that no change to MABS would achieve what you are seeking ?
> 
> Lack of heif support seems to be a long overdue shortcoming in ffmpeg, based on TRAC comments.

It's funny (or not) that ffmpeg supports so called "avif" but no "heic". It's just a container!
No big deal though, things are just starting to get complicated as some people already *removed* jpeg-xl.

hydra3333:
Yes, things seem a tad peculiar. 
For better or worse, I'd always thought of ffmpeg as a mainstream video processing product and so was unsurprised to note some deficiencies around image processing.

gitoss:
> hi, is it possible to add libheif to this script? https://github.com/strukturag/libheif

Using media-autobild's msys shell, I was able to compile libheif with all HEVC & AV1 encoders/decoders I want to use - there are no Windows binaries around the web for that. 

[...]

PR #2796, opened 2024-11-18
Add libheif (plus some small fixes)

L4cache:
Implements #2516
It's not perfect but at least most important functions are working, and I don't think I can fix the remaining issues.


build/media-suite_compile.sh
```diff
@@ -2070,6 +2070,7 @@ if [[ $ffmpeg != no ]] && enabled libxeve &&
     mv -f "$LOCALDESTDIR"/lib/xeve/libxeve.a "$LOCALDESTDIR"/lib/libxeve.a
     mv -f "$LOCALDESTDIR"/lib/libxeve.dll.a "$LOCALDESTDIR"/lib/libxeve.dll.a.dyn
     do_checkIfExist
+    rmdir "$LOCALDESTDIR/lib/xeve" > /dev/null 2>&1
```
1480c1, line 2073: Can this have a note about why this is explicitly done here? Also, can you move this right next to the `mv` to make it clear that the reason this dir is not needed is that the stuff inside has been moved out?

build/media-suite_compile.sh
```diff
@@ -2082,6 +2083,7 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
     mv -f "$LOCALDESTDIR"/lib/xevd/libxevd.a "$LOCALDESTDIR"/lib/libxevd.a
     mv -f "$LOCALDESTDIR"/lib/libxevd.dll.a "$LOCALDESTDIR"/lib/libxevd.dll.a.dyn
     do_checkIfExist
+    rmdir "$LOCALDESTDIR/lib/xevd" > /dev/null 2>&1
```
1480c1, line 2086: ditto.

build/media-suite_compile.sh
```diff
@@ -2424,6 +2426,75 @@ if [[ $ffmpeg != no ]]; then
...
+
+_check=(bin-video/heif-{dec,enc,info,thumbnailer}.exe)
+if [[ $libheif = y ]] &&
+    do_vcs "$SOURCE_REPO_LIBHEIF"; then
+    do_uninstall "${_check[@]}"
+
+    sed -i 's/find_package(vvdec 2.3.0)/find_package(vvdec 3.0.0)/' CMakeLists.txt
+    sed -i 's/find_package(Doxygen)/#/' CMakeLists.txt # no configurable option?
+    sed -i 's/find_package(Brotli)/#/' CMakeLists.txt # linking difficulties
+    sed -i 's/find_package(TIFF)/#/' heifio/CMakeLists.txt # configure & linking difficulties
```
1480c1, line 2448: I would prefer to fix these upstream if possible. If not, or if upstream rejects changes to fix these, then I am fine with adding them here.

build/media-suite_compile.sh
```diff
@@ -2424,6 +2426,75 @@ if [[ $ffmpeg != no ]]; then
...
+
+    extracflags=()
+    extracommands=(-DWITH_HEADER_COMPRESSION=ON -DWITH_UNCOMPRESSED_CODEC=ON)
+    pc_exists "libde265" &&
+        extracommands+=(-DWITH_LIBDE265=ON -DWITH_LIBDE265_PLUGIN=OFF) &&
```
1480c1, line 2453: I haven't looked at upstream yet, what's the deal with the `PLUGIN=OFF` stuff?

build/media-suite_compile.sh
```diff
@@ -2424,6 +2426,75 @@ if [[ $ffmpeg != no ]]; then
...
+    # linking difficulties
+    pc_exists "rav1e" &&
+        extracommands+=(-DWITH_RAV1E=OFF -DWITH_RAV1E_PLUGIN=OFF)
+    pc_exists "libavcodec" "libavutil" &&
+        extracommands+=(-DWITH_FFMPEG_DECODER=OFF -DWITH_FFMPEG_DECODER_PLUGIN=OFF)
+
+    CFLAGS+=" ${extracflags[@]}" CXXFLAGS+=" ${extracflags[@]}" \
+        do_cmakeinstall video -DBUILD_TESTING=OFF -DWITH_GDK_PIXBUF=OFF "${extracommands[@]}"
+    do_checkIfExist
+    rmdir "$LOCALDESTDIR/lib/libheif" > /dev/null 2>&1
```
1480c1, line 2495: Note on this too?

build/media-suite_compile.sh
L4cache: What note? Like a comment? Shall it be about `rmdir` only or including why do `mv` as well?

build/media-suite_compile.sh
L4cache: Some of the codecs will be built as a plugin dll, even with BUILD_SHARED_LIBRARIES=OFF. But the plugin themselves don't have extra dependencies though.

build/media-suite_compile.sh
1480c1: Yeah, a comment as why this is fine, something like `# delete empty folder`. Just so later on, if we see this, we know that we are assuming the folder is empty due to the previous `mv`s.

build/media-suite_compile.sh
1480c1: Understood

build/media-suite_compile.sh
L4cache: Man, I don't think they are going to be fixed anytime soon, maybe we can merge it for now, or transfer them into patches?

1480c1: I would prefer transferring to patches since it makes it easier for me to know if it breaks or is no longer needed.

L4cache: Can you push to the branch and patches repo when you decide to merge this PR? Or shall I do it and also open a PR in the patches repo?

1480c1: I would prefer that you do it so the authorship of the changes is preserved

L4cache:
Where's that comment suggesting a shared option?

1480c1:
> Where's that comment suggesting a shared option?

Idk, I think it was something about making it so that most of the binaries are shared or something since they link a lot of libraries?

L4cache:
And the patch...

PR #2798, opened 2024-11-19
Misc improvements

woot000:
* Download `media-suite_deps.sh` if not found
* Update upx, gnutls, libidn2, libpsl, and openh264 to their newest versions
* Make the msys2 package update interval configurable in .ini

I set up `pkgUpdateTime` to not require user input because the default of one day is good enough, and I don't think it would be necessary to change it unless a user *needed* more frequent package updates

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -181,9 +181,9 @@ if [[ ! -z $_reenable_av1an ]] && [[ $bits = 64bit ]]; then
 fi
 
 if [[ $packing = y &&
-    ! "$(/opt/bin/upx -V 2> /dev/null | head -1)" = "upx 4.2.2" ]] &&
-    do_wget -h 141e7cd8d009b827590662b482f1ae2f1dda17cf446a5651078235efb1429c59 \
-        "https://github.com/upx/upx/releases/download/v4.2.2/upx-4.2.2-win32.zip"; then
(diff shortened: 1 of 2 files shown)


build/media-suite_compile.sh
```diff
@@ -402,24 +402,25 @@ if [[ $mediainfo = y || $bmx = y || $curl != n ]]; then
...
+        do_wget -h 4c21a791b610b9519b9d0e12b8097bf2f359b12f8dd92647611a929e6bfd7d64 \
+        "https://ftp.gnu.org/gnu/libidn/libidn2-2.3.7.tar.gz"; then
         do_uninstall "${_check[@]}"
+        do_pacman_install gtk-doc
         [[ $standalone == y ]] || sed -ri 's|(bin_PROGRAMS = ).*|\1|g' src/Makefile.am
         # unistring also depends on iconv
         grep_or_sed '@LTLIBUNISTRING@ @LTLIBICONV@' libidn2.pc.in \
             's|(@LTLIBICONV@) (@LTLIBUNISTRING@)|\2 \1|'
-	do_autoreconf   
+	    AUTOPOINT=true do_autoreconf
```
1480c1, line 414: why does the indentation here look weird?

woot000: Looks fine on my end
{034C0CDA-066C-4585-B6A5-9951FFFA30E9}

1480c1: lol.

PR #2817, opened 2025-01-18
prevent superfluous clang installation during FFmpeg compilation

L4cache:
Although cuda-llvm is included in default ffmpeg options, when using custom ffmpeg options, if the cuda-llvm is not explicitly disabled (it should be implicitly disabled by --disable-autodetect which is default option too) it will move on to install the clang package.

If a user, now, after the removal of clang from default packages, and is using GCC toolchain, "implicitly disabled" cuda-llvm by remove or comment out the line in ffmpeg_options, the clang package will be installed without any actual use.

I'm not entirely sure what this part of the code is intended for, maybe this is not the best way to prevent it.


1480c1:
This was during the time when cuda-llvm was introduced, and we were on cuda-nvcc for a long time, and most people didn't know to switch to cuda-llvm.

PR #2833, opened 2025-02-16
Misc changes

woot000:
A bunch of small edits as follows:
* Update python to 3.12.9 and gnutls to 3.8.9
* Remove CLANG32 support, since it has been officially removed from msys2 for a couple months now
* Fix building ffmpeg with gmp using CLANG64, since it isn't automatically installed as it is when building with gcc
* Remove a bunch of old/redundant code
* Tweak some lines of code to be more consistent with each other
* Move some code to `media-suite_helper.sh` and some git urls to `media-suite_deps.sh`


build/media-suite_helper.sh
```diff
@@ -2224,76 +2220,47 @@ create_cmake_toolchain() {
...
+    # current abspath root
+    local _root
+    _root=$(cygpath -m "$LOCALDESTDIR")
+    # find .pc files with Windows abspaths
+    grep -ElZR "${_keys}=[^/$].*" "$LOCALDESTDIR"/lib/pkgconfig | \
+        # find those with a different abspath than the current
+        xargs -0r grep -LZ "$_root" | \
+        # replace with current abspath
+        xargs -0r sed -ri "s;${_keys}=.*$LOCALDESTDIR;\1=$_root;g"
+    unset _keys _root
```
1480c1, line 2237: Vaguely sure this line is not needed and has no effect overall since these are declared as local to the function and should be automatically cleaned up when bash exits the function

media-autobuild_suite.bat
```diff
@@ -1549,12 +1548,12 @@ if [0]==[%ccacheINI%] (
...
     echo. Use ccache when compiling?
-    echo. Experimental.
-    echo. Speeds up rebuilds and recompilations, but requires the files to be
-    echo. compiled at least once before any effect is seen
     echo. 1 = Yes
     echo. 2 = No
     echo.
+    echo. Experimental. Speeds up rebuilds and recompilations, but requires the files
+    echo. to be compiled at least once before any effect is seen
+    echo.
```
1480c1, line 1556: I think experimental can be removed from this, as I've been compiling with ccache ever since this was introduced and I haven't had any major issues with it.

media-autobuild_suite.bat
```diff
@@ -2072,9 +2084,9 @@ echo.---------------------------------------------------------------------------
 echo.install %1 bit compiler
 echo.-------------------------------------------------------------------------------
 if %CC%==clang (
-    if "%1"=="32" (
-        set prefix=mingw-w64-clang-i686-
-    ) else set prefix=mingw-w64-clang-x86_64-
+    if "%1"=="64" (
+        set prefix=mingw-w64-clang-x86_64-
+    )
```
1480c1, line 2089:
```suggestion
    if "%1"=="64" set prefix=mingw-w64-clang-x86_64-
```

Tbh, I'm pretty sure this if check can be deleted since we check that `CC==clang && build32==yes` is an invalid state, so `call :getmingw 32` should never be called when `CC==clang`.

media-autobuild_suite.bat
```diff
@@ -1900,7 +1912,7 @@ set compileArgs=--cpuCount=%cpuCount% --build32=%build32% --build64=%build64% ^
     if %build64%==yes (
         if %CC%==clang ( set "MSYSTEM=CLANG64" ) else set "MSYSTEM=MINGW64"
     ) else (
-        if %CC%==clang ( set "MSYSTEM=CLANG32" ) else set "MSYSTEM=MINGW32"
+        if %CC%==gcc set "MSYSTEM=MINGW32"
     )
```
1480c1, line 1916: Same here regarding `CC==clang && build32=yes` being an invalid state, so assuming `build64 != yes`, then `build32` must equal yes, and thus `CC` must be `gcc` (at least in our case).

1480c1:
Thanks

PR #2855, opened 2025-03-19
Remove patch to compactify extralibs

GyanD:
ffmpeg now passes objects list to linker from a file since https://github.com/FFmpeg/FFmpeg/commit/9e857e1f8aeeb655adb9d09bf53add26a27092e2

so this step should no longer be necessary.


1480c1:
Thanks

PR #2900, opened 2025-04-30
[mpv] add  libnpp build flags into Meson configuration

okazunori2013:
When libnpp is enabled, the build fails, so we fixed it by specifying the path.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2841,6 +2841,12 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
         if ! mpv_disabled manpage-build || mpv_enabled html-build; then
             do_pacman_install python-docutils
         fi
+        if enabled libnpp && [[ -n "$CUDA_PATH" ]]; then
+            mpv_cflags=("-I$(cygpath -sm "$CUDA_PATH")/include")
+            mpv_ldflags=("-L$(cygpath -sm "$CUDA_PATH")/lib/x64")
+            MPV_OPTS["c_args"]="${mpv_cflags[*]}"
+            MPV_OPTS["c_link_args"]="${mpv_ldflags[*]}"
+        fi        
         mpv_enabled pdf-build && do_pacman_install python-rst2pdf
 
         [[ -f mpv_extra.sh ]] && source mpv_extra.sh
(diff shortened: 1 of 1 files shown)


build/media-suite_compile.sh
```diff
@@ -2841,6 +2841,12 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
         if ! mpv_disabled manpage-build || mpv_enabled html-build; then
             do_pacman_install python-docutils
         fi
+        if enabled libnpp && [[ -n "$CUDA_PATH" ]]; then
+            mpv_cflags=("-I$(cygpath -sm "$CUDA_PATH")/include")
+            mpv_ldflags=("-L$(cygpath -sm "$CUDA_PATH")/lib/x64")
+            MPV_OPTS["c_args"]="${mpv_cflags[*]}"
+            MPV_OPTS["c_link_args"]="${mpv_ldflags[*]}"
```
1480c1, line 2848: I would be wary of this based on looking at meson's documentation https://mesonbuild.com/Builtin-options.html#compiler-options
> c_args free-form comma-separated list

build/media-suite_compile.sh
```diff
@@ -2841,6 +2841,12 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
         if ! mpv_disabled manpage-build || mpv_enabled html-build; then
             do_pacman_install python-docutils
         fi
+        if enabled libnpp && [[ -n "$CUDA_PATH" ]]; then
+            mpv_cflags=("-I$(cygpath -sm "$CUDA_PATH")/include")
```
1480c1, line 2845: Just realized you are calling cygpath and not pkg-config here. Since these are just single strings, there's no need to make it an array.

build/media-suite_compile.sh
```diff
@@ -2841,6 +2841,12 @@ if [[ $mpv != n ]] && pc_exists libavcodec libavformat libswscale libavfilter; t
...
+            mpv_ldflags=("-L$(cygpath -sm "$CUDA_PATH")/lib/x64")
+            MPV_OPTS["c_args"]="$(IFS=,; echo "${mpv_cflags[*]}")"
```
1480c1, line 2847: The reason I mentioned these lines was in consideration of someone having c flags already here. This would completely override them.

1480c1:
Added in https://github.com/m-ab-s/media-autobuild_suite/commit/fab6ce6780c06a6ad5055ce17991227739ef6ace

PR #2910, opened 2025-05-08
Add openapv (enable through FFmpeg options only) and misc change

L4cache:
Currently the static link doesn't work.
Maybe better wait sometime to see if it will be fixed?

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2215,6 +2215,16 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
     do_checkIfExist
 fi
 
+_check=(bin-video/oapv_app_{enc,dec}.exe oapv/oapv.h oapv/liboapv.a oapv.pc)
+if [[ $ffmpeg != no ]] && enabled liboapv &&
+    do_vcs "$SOURCE_REPO_OPENAPV"; then
+    do_uninstall "${_check[@]}"
+    do_cmakeinstall video -DOAPV_BUILD_SHARED_LIB=OFF
+    # patch the lib path to actual subdirectory it installed to
+    sed -i 's|Libs: -L${libdir} -loapv|Libs: -L${libdir}/oapv -loapv|' "$LOCALDESTDIR"/lib/pkgconfig/oapv.pc
+    do_checkIfExist
+fi
+
 _check=(avisynth/avisynth{,_c}.h
         avisynth/avs/{alignment,arch,capi,config,cpuid,minmax,posix,types,win,version}.h)
 if [[ $ffmpeg != no ]] && enabled avisynth &&
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -2191,33 +2191,27 @@ else
     pc_exists libvvdec || do_removeOption "--enable-libvvdec"
 fi
 
-_check=(bin-video/xeve_app.exe xeve/xeve{,_exports}.h libxeve.a xeve.pc)
+_check=(bin-video/xeve_app.exe xeve/xeve{,_exports}.h xeve/libxeve.a xeve.pc)
 if [[ $ffmpeg != no ]] && enabled libxeve &&
     do_vcs "$SOURCE_REPO_XEVE"; then
     do_uninstall bin-video/libxeve.dll lib/libxeve.dll.a.dyn "${_check[@]}"
     sed -i 's/-Werror //' CMakeLists.txt
-    do_cmakeinstall video
-    # no way to disable shared lib building in cmake
-    # move the static library out from subfolder to make ffmpeg configure find it easier
-    mv -f "$LOCALDESTDIR"/lib/xeve/libxeve.a "$LOCALDESTDIR"/lib/libxeve.a
+    do_cmakeinstall video # no way to disable shared lib building in cmake
     mv -f "$LOCALDESTDIR"/lib/libxeve.dll.a "$LOCALDESTDIR"/lib/libxeve.dll.a.dyn
-    # delete the now empty subfolder
-    rmdir "$LOCALDESTDIR/lib/xeve" > /dev/null 2>&1
+    # patch the lib path to actual subdirectory it installed to
+    sed -i 's|Libs: -L${libdir} -lxeve|Libs: -L${libdir}/xeve -lxeve|' "$LOCALDESTDIR"/lib/pkgconfig/xeve.pc
     do_checkIfExist
 fi
 
-_check=(bin-video/xevd_app.exe xevd/xevd{,_exports}.h libxevd.a xevd.pc)
+_check=(bin-video/xevd_app.exe xevd/xevd{,_exports}.h xevd/libxevd.a xevd.pc)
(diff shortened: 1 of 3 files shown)


L4cache:
Static link is fixed

L4cache:
@1480c1 What's your thought?

1480c1:
I'm first working on fixing any other obvious issue before I introduce a new library, but if I don't get any progress done by tomorrow, I'll probably just merge this.

build/media-suite_compile.sh
```diff
@@ -2221,6 +2221,21 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
     do_checkIfExist
 fi
 
+_check=(bin-video/oapv_app_{enc,dec}.exe oapv/oapv{,_exports}.h liboapv.a oapv.pc)
+if [[ $ffmpeg != no ]] && enabled liboapv &&
+    do_vcs "$SOURCE_REPO_OPENAPV"; then
+    do_uninstall bin-video/liboapv.dll lib/liboapv.dll.a.dyn "${_check[@]}"
+    do_cmakeinstall video -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_SYSTEM_PROCESSOR=x86_64
+    # no way to disable shared lib building in cmake
+    # move the libraries out from subfolder to make ffmpeg configure find it easier
```
1480c1, line 2230: This fact really sucks :(

build/media-suite_compile.sh
```diff
@@ -2221,6 +2221,21 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
...
+    mv -f "$LOCALDESTDIR"/lib/oapv/liboapv.a "$LOCALDESTDIR"/lib/liboapv.a
```
1480c1, line 2231: in theory, as long as their pkg-config file states that this directory as the library location, then it should be okay to not have to move it

build/media-suite_compile.sh
```diff
@@ -2221,6 +2221,21 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
...
+    mv -f "$LOCALDESTDIR"/lib/oapv/import/liboapv.dll.a "$LOCALDESTDIR"/lib/liboapv.dll.a.dyn
```
1480c1, line 2232: ~~this can be covered by a hide_all_shared_libs or whatever the function is called~~

just noticed the special `/import/` part of the path, not sure what to make of it yet.

build/media-suite_compile.sh
```diff
@@ -2221,6 +2221,21 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
...
+    # delete the now empty subfolder
+    rmdir "$LOCALDESTDIR/lib/oapv/import" > /dev/null 2>&1
+    rmdir "$LOCALDESTDIR/lib/oapv" > /dev/null 2>&1
```
1480c1, line 2235: Okay, wait, I just started reading upstream's cmake and pkg-config file and I can say I'm not entirely sure what their goals are.

media-autobuild_suite.bat
```diff
@@ -409,8 +409,10 @@ if [0]==[%libheifINI%] (
     echo.
     echo. Will use available encoders and decoders supported by libheif.
     echo. If not found, built libheif will lack the corresponding encode/decode ability.
-    echo. Additionally libde265 will be built.
-    echo. dec265 of libde265 being built depends on "standalone=y" and is always static.
+    echo.
+    echo. Additionally libde265 will always be built.
+    echo. dec265 executable from libde265 being built depends on "standalone=y"
+    echo. and is always static.
```
1480c1, line 415: If this commit is split off into its own PR, I can easily merge it.

build/media-suite_compile.sh
L4cache: So we can patch the pkgconf file (or fix it upstream) instead of moving files?

media-autobuild_suite.bat
L4cache: No hurry for this trivial stuff, I guess...

build/media-suite_compile.sh
1480c1: Would definitely prefer to fix upstream, but a patch is fine too

1480c1: Can you post the content of the resulting pkg-config file when you run mabs?

L4cache:
```
# --- oapv.pc.in file ---
prefix=C:/mabs/local64 
exec_prefix=C:/mabs/local64/C:/mabs/local64/bin-video
libdir=C:/mabs/local64/lib
includedir=C:/mabs/local64/include/oapv

Name: oapv
Description: Advanced Professional Video Codec

Version: 0.1.13

Requires:
Libs: -L${libdir} -loapv
Libs.private: -L${libdir}/oapv -lm

Cflags: -I${includedir}
Cflags.private: -DOAPV_STATIC_DEFINE
```

L4cache: Add /oapv subfolder into `Libs` field, and FFmpeg or more accurately pkgconfig is able to find the library no problem...

L4cache:
Now it has option to disable shared lib

build/media-suite_compile.sh
1480c1: I agree on the exec_prefix being weird, I can guess the issue is that we set bindir like `bindir="-DCMAKE_INSTALL_BINDIR=$LOCALDESTDIR/bin-$1"`. I don't know if it will break other cmake projects if we just set it to `bindir="-DCMAKE_INSTALL_BINDIR=bin-$1"`

build/media-suite_compile.sh
```diff
@@ -2221,6 +2221,20 @@ if [[ $ffmpeg != no ]] && enabled libxevd &&
     do_checkIfExist
 fi
 
+_check=(bin-video/oapv_app_{enc,dec}.exe oapv/oapv{,_exports}.h liboapv.a oapv.pc)
+if [[ $ffmpeg != no ]] && enabled liboapv &&
+    do_vcs "$SOURCE_REPO_OPENAPV"; then
+    do_uninstall bin-video/liboapv.dll lib/liboapv.dll.a.dyn "${_check[@]}"
+    do_cmakeinstall video -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DOAPV_BUILD_SHARED_LIB=OFF
```
1480c1, line 2228: Are you sure `CMAKE_SYSTEM_NAME` and `CMAKE_SYSTEM_PROCESSOR` are necessary here? I believe they should be automatically set by cmake itself?

L4cache: I was sure when I wrote this, not entirely sure now, but I guess so...

1480c1: Can you check if these variables are set in the cmake cache if you don't have them set here?

L4cache: The build was failing without these.
But it's now fixed with https://github.com/AcademySoftwareFoundation/openapv/commit/649f81eca5162aa86cddbcd9f28f26395e8e300a

PR #2917, opened 2025-05-14
Disable (legacy) jpegli inside libjxl

gitoss:
The jpegli source has been branched from libjxl - according to the libjxl CMakeLists.txt, the default is libjpeg-turbo from the third_party subrepo (which is pulled by mabs).

I've used -DJPEGXL_ENABLE_JPEGLI=OFF in a libjxl_extra.sh for quite some time, but always had to remove the tools check for cjpegli/djpegli from media-suite_compile.sh

Moving jpegli completely outside libjxl and using a dependency seems to have stalled, but nevertheless jpegli development doesn't happen inside libjxl anymore. If the jpegli tools should be a built-in feature of mabs (probably not), it should be separate and use https://github.com/google/jpegli

If mabs should require a jpeg lib in local, it should be separate from the legacy 'just happens to be there' jpegli inside libjxl, too.


1480c1:
Thanks

PR #2923, opened 2025-05-18
Add support for AudioToolbox via AudioToolboxWrapper

KokonoeTooru:
Sed code is based of  [M05QU170](https://github.com/user-attachments/files/19558379/media-suite_compile-audiotoolbox.patch) patch.

I'm not really sure if this is the right way to do this....
Also current implementation might not [decode probably](https://github.com/AnimMouse/QTFiles/tree/v12.10.11?tab=readme-ov-file#icudt62dll-dummy-file), though my limited testing show no difference between the original `icudt62.dll` and the dummy one.

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1182,6 +1182,27 @@ if [[ $ffmpeg != no ]] && enabled liblc3 &&
     do_checkIfExist
 fi
 
+_check=(bin/atw_ldwrapper libAudioToolboxWrapper.a
+    bin-video/{ASL,CoreAudioToolbox,CoreFoundation,icudt62,libdispatch,libicuin,libicuuc,objc}.dll)
+if [[ $ffmpeg != no ]] && enabled audiotoolbox &&
+    do_vcs "$SOURCE_REPO_AUDIOTOOLBOX"; then
+    _qtfiles_url="https://github.com/AnimMouse/QTFiles/releases/download/v12.10.11"
+    do_uninstall "${_check[@]}"
+    if [[ $build64 = yes ]]; then
+        do_wget -c -r -q "${_qtfiles_url}/QTfiles64.7z" QTfiles64.7z
+        do_install QTfiles64/*.dll bin-video
+        rm -rf QTfiles64/ QTfiles64.7z
+    fi
+    if [[ $build32 = yes ]]; then
+        do_wget -c -r -q "${_qtfiles_url}/QTfiles.7z" QTfiles.7z
+        do_install QTfiles/*.dll bin-video
+        rm -rf QTfiles/ QTfiles.7z
+    fi
+    do_cmakeinstall
+    do_checkIfExist
+    unset _ver _qtfiles_url
+fi
+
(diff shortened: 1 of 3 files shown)


build/media-suite_compile.sh
```diff
@@ -1182,6 +1182,27 @@ if [[ $ffmpeg != no ]] && enabled liblc3 &&
...
+    do_vcs "$SOURCE_REPO_AUDIOTOOLBOX"; then
+    _qtfiles_url="https://github.com/AnimMouse/QTFiles/releases/download/v12.10.11"
+    do_uninstall "${_check[@]}"
+    if [[ $build64 = yes ]]; then
+        do_wget -c -r -q "${_qtfiles_url}/QTfiles64.7z" QTfiles64.7z
+        do_install QTfiles64/*.dll bin-video
+        rm -rf QTfiles64/ QTfiles64.7z
+    fi
+    if [[ $build32 = yes ]]; then
+        do_wget -c -r -q "${_qtfiles_url}/QTfiles.7z" QTfiles.7z
```
1480c1, line 1197: Please change the do_wget lines to remove the `-c` and add a hash to check against with the `-h` argument, preferable a sha256 one. This can help weed out issues regarding internet connection and avoid redownloading if the file is already present. This can help in cases where a user nukes their local64/local32 folder but kept the build folder intact.

build/media-suite_compile.sh
```diff
@@ -1182,6 +1182,27 @@ if [[ $ffmpeg != no ]] && enabled liblc3 &&
...
+        do_install QTfiles/*.dll bin-video
+        rm -rf QTfiles/ QTfiles.7z
+    fi
+    do_cmakeinstall
+    do_checkIfExist
+    unset _ver _qtfiles_url
```
1480c1, line 1203: `_ver` doesn't seem to be used here.

build/media-suite_compile.sh
```diff
@@ -2465,6 +2486,10 @@ if [[ $ffmpeg != no ]]; then
 
         enabled libvvdec && grep_and_sed FF_PROFILE libavcodec/libvvdec.c 's/FF_PROFILE/AV_PROFILE/g'
 
+        # Bypass ffmpeg check for audiotoolbox
+        enabled audiotoolbox && do_addOption --extra-libs=-lAudioToolboxWrapper && do_addOption --disable-outdev=audiotoolbox &&
+            sed -ri "s/check_apple_framework AudioToolbox/check_apple_framework/g" /build/ffmpeg-git/configure
```
1480c1, line 2491:
```suggestion
        enabled audiotoolbox && do_addOption --extra-libs=-lAudioToolboxWrapper && do_addOption --disable-outdev=audiotoolbox &&
            sed -ri "s/check_apple_framework AudioToolbox/check_apple_framework/g" configure
```

You should already be in `/build/ffmpeg-git` by now, so this full path will only induce confusion.

1480c1 (commented):
Please rework the do_wget parts to actually take advantage of being able to keep QTfiles*.7z between builds. This means not rm'ing them etc

PR #2950, opened 2025-06-07
Fixed Allow *_extra.sh scripts to print info.

smiRaphi:
I just added a specific new option for the logging in extra_script so nothing else should be affected this time.

### DIFF
--- a/build/media-suite_helper.sh
+++ b/build/media-suite_helper.sh
@@ -1512,11 +1512,13 @@ zip_logs() {
 }
 
 log() {
-    local errorOut=true quiet=false ret OPTION OPTIND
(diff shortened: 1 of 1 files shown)


diegocr:
Hi guys, i've been using this `curl_extra.sh` script _for ages_ which is no longer working, is this PR the culprit?

```
_pre_vcs() {
    # ref=ec4eec222b4ccc48c3f711b3cad31d9026919d46

    ref=$(cd "${REPO_DIR}" && git log -n 1 --pretty=format:"%H")
}
```

1480c1:
It shouldn't be. You can try by using git checkout to go to the revision before this pr and see

PR #2969, opened 2025-07-09
Add option to keep legacy configure arguments

smiRaphi:
It can only be enabled by editing the .ini directly so users normally shouldn't be able to break stuff with it.


L4cache:
There are more places that clone ffmpeg

smiRaphi:
@L4cache where? I know that there are some local clones but those don't really need changes.

smiRaphi:
@1480c1 is there a reason as to why this isn't being merged?

1480c1:
At some point, it was because you asked a question to @L4cache. Afterward, it's because I forgot.

PR #3031, opened 2025-11-02
Create ffmpeg_options.txt

kahfi112:


1480c1:
The options file is not supposed to be checked into the repo.

PR #3044, opened 2025-12-15
add SVT-JPEG-XS

L4cache:
Is it better installed into bin-video or not?


1480c1:
> Is it better installed into bin-video or not?

Based on looking at other image centric formats like libjxl, keep in global

1480c1:
Visually lgtm, going to go ahead and merge and test after

PR #3060, opened 2026-01-28
Install libid3tag for sox

MarcroSoft:
Sox uses libid3tag to read tags from mp3 files, so install it along with libmad.
Please do not merge this until msys2 creates a static libid3tag package, see this issue:
https://github.com/msys2/MINGW-packages/issues/27587


MarcroSoft:
This is now ready to be reviewed and merged.

PR #3089, opened 2026-03-12
temp mediainfo fix

God-damnit-all:
Recently the MediaArea repos all had significant cmake overhauls that break m-ab-s, this sets the commits to right before those breaking changes until a proper fix is implemented.


L4cache:
I don't see problem now, what *was* the problem?

God-damnit-all:
> I don't see problem now, what _was_ the problem?

cmake issues, looks like they pushed a new set of commits on their repos shortly after I made this PR. They'd been broken for a few days so I thought I'd submit my fix, but wouldn't you know they fixed it immediately afterward.

PR #3122, opened 2026-05-18
zlib-ng: reduce static minizip-ng build fail to warning

### LINKED ISSUE #3115: mingw64 clang64 minizip-ng configure failed
15:08:32 ┌ zlib-ng git  .............................. [Recently updated]
15:08:32 ├ Running uninstall...
15:08:32 ├ Running cmake...
15:08:39 ├ Running build...
15:08:39 ├ Running install...
15:08:39   Running git clone for minizip-ng ...
15:08:45   Running git update for minizip-ng ...
15:08:45 ┌ minizip-ng git  ........................... [Recently updated]
15:08:45 ├ Running cmake...
Likely error (tail of the failed operation logfile):
     "D:/ffmpeg-autobuild-clang/msys64/clang64/lib/cmake/zstd/zstdTargets.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  D:/ffmpeg-autobuild-clang/msys64/clang64/lib/cmake/zstd/zstdConfig.cmake:32 (include)
  CMakeLists.txt:402 (find_package)

-- Configuring incomplete, errors occurred!
cmake failed. Check D:/ffmpeg-autobuild-clang/build/minizip-ng-git/build-64bit/ab-suite.cmake.log

[ab-suite.cmake.log](https://github.com/user-attachments/files/27317390/ab-suite.cmake.log)

Highlandword9:
Fixes #3115

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -288,6 +288,10 @@ else
             do_cmakeinstall global -DZLIB_COMPAT=ON -DWITH_GTEST=OFF -DZLIB_ENABLE_TESTS=OFF
             if [[ $standalone = y ]] &&
                 do_vcs "$SOURCE_REPO_MINIZIPNG"; then
+                # Only demote to a warning if shared mode is NOT explicitly requested
+                if ! [[ ${shared:-n} == y || ${enable_shared:-n} == y || ${BUILD_SHARED_LIBS:-OFF} == ON ]]; then
+                    sed -i 's;message(FATAL_ERROR "The imported target;message(WARNING "The imported target;' "$MINGW_PREFIX"/lib/cmake/zstd/zstdTargets.cmake
+                fi
                 do_cmakeinstall global -DMZ_BUILD_TESTS=ON
             fi
             do_checkIfExist


build/media-suite_compile.sh
```diff
@@ -288,6 +288,10 @@ else
             do_cmakeinstall global -DZLIB_COMPAT=ON -DWITH_GTEST=OFF -DZLIB_ENABLE_TESTS=OFF
             if [[ $standalone = y ]] &&
                 do_vcs "$SOURCE_REPO_MINIZIPNG"; then
+                # Only demote to a warning if shared mode is NOT explicitly requested
+                if ! [[ ${shared:-n} == y || ${enable_shared:-n} == y || ${BUILD_SHARED_LIBS:-OFF} == ON ]]; then
```
1480c1, line 292: None of these variables are used/present in the suite.

Generally, when we are determining to build static or shared, we check `$ffmpeg = sharedlibs`, however, in this case, I would prefer to just unconditionally run the sed command, similar to how we do for libtiff.

Highlandword9: Ah, ok. I remembered them from a project I did a few years ago with cmake, couldn't find them here but to be safe I put them in.

I was being too cautious it seems ;]
If unconditional is fine that sed should work.

Issue #3126, opened 2026-05-25
[mingw64] mpv build fails with libarchive and pacman zlib


ultrasound1372:
FFmpeg builds fine, but MPV fails to build with an entire cascade of linker errors which appear to be coming from zlib via libarchive. As is recommended by the suite itself, I use the zlib package from pacman.
``` bash
┌ mpv git  ............................................... [Updates found]
├ Running meson...
├ Running build...
Likely error (tail of the failed operation logfile):
C:/tools/mabs/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/tools/mabs/msys64/ucrt64/lib/libarchive.a(archive_read_support_format_zip.c.obj):(.text$real_crc32+0x3): undefined reference to `__imp_crc32'
C:/tools/mabs/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/tools/mabs/msys64/ucrt64/lib/libarchive.a(archive_read_support_format_zip.c.obj):(.text$zip_read_data_deflate.isra.0+0x5a): undefined reference to `__imp_inflateReset'
C:/tools/mabs/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/tools/mabs/msys64/ucrt64/lib/libarchive.a(archive_read_support_format_zip.c.obj):(.text$zip_read_data_deflate.isra.0+0x12d): undefined reference to `__imp_inflate'
C:/tools/mabs/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/tools/mabs/msys64/ucrt64/lib/libarchive.a(archive_read_support_format_zip.c.obj):(.text$zip_read_data_deflate.isra.0+0x394): undefined reference to `__imp_inflateInit2_'
[...]

PR #3127, opened 2026-05-29
Patch all providers in fix_impsyms

GyanD:
The user may have residual and differing versions of a library and we can't always predict which version a package may use. So patch all found providers.

The proper fix is to run pkg config on all entries in Libs and identify specific location of each entry.


GyanD:
Should fix #3126

Highlandword9:
works a charm!

PR #3147, opened 2026-07-11
Add multiple build options

### LINKED ISSUE #3021: Request: Support for mpeghdec
mpeghdec has been added to ffmpeg and it would be really useful if this could be added within the build system.

ambitiuswisdom:
## Summary

This PR ~adds configurable CPU targeting for 64-bit builds and~ expands support for several optional FFmpeg dependencies:

- ~Add selectable x86-64 CPU targets~.
- Restore static libssh support.
- Add an optional 10-bit davs2 build.
- Add Fraunhofer libmpeghdec support.

### ~Configurable CPU target~ (Removed)

~Add a `targetPlatform` option for selecting `x86-64`, `x86-64-v2`,
`x86-64-v3`, `x86-64-v4`, or `native`.~

~This allows users to balance binary compatibility and performance instead
of forcing every 64-bit build to use the same generic CPU baseline. Newer
targets let the compiler use instruction sets available on modern CPUs,
which may improve performance in compute-intensive multimedia workloads.~

~The selected target is applied consistently to C/C++ and Rust code.
For combined 32/64-bit builds, the 32-bit output remains generic to preserve
compatibility.~

### Restore libssh support

Re-enable FFmpeg's `libssh` integration using the MinGW-w64 package.

The generated pkg-config metadata is supplemented with the static dependencies required by libssh, including OpenSSL, zlib, Winsock, IP Helper API, and pthread. 

### Optional 10-bit davs2

Add an opt-in 10-bit davs2 decoder using the unofficial
`saindriches/davs2` fork. The official `pkuvcl/davs2` repository currently
remains the source for the existing 8-bit option and is not replaced.

The fork is explicitly presented as unofficial because it is maintained
outside the upstream davs2 project. Users should therefore be aware that
its maintenance status, compatibility, and review process may differ from
the official implementation.

### Add libmpeghdec(Fixes #3021)

Add build support for Fraunhofer IIS's MPEG-H decoder:

- Build and install the static decoder library.
- Enable the decoder and UI manager executables for standalone builds.
- Add the required `MPEGHDEC_STATIC` pkg-config flag.
- ~Apply a small MinGW compatibility adjustment to avoid duplicate `SHORT` overloads under the LLP64 data model.(upstream issue https://github.com/Fraunhofer-IIS/mpeghdec/issues/17)~

### DIFF
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1298,6 +1298,11 @@ if [[ $ffmpeg != no ]] && enabled libmpeghdec &&
         extracommands=(-Dmpeghdec_BUILD_BINARIES=OFF -Dmpeghdec_BUILD_UIMANAGER=OFF)
     fi
     do_cmakeinstall "${extracommands[@]}" -DCMAKE_INSTALL_DATAROOTDIR=lib
+    # Avoid bundled FDK symbol collisions with libfdk-aac.
+    if enabled libfdk-aac; then
+        prefix_archive_symbols "$LOCALDESTDIR/lib/libmpeghdec.a" \
+            mpeghdec_private_ '^_?(mpeghdecoder_|mpegh_UI_)'
+    fi
     [[ $standalone = y ]] &&
         do_install bin/{mpeghDecoder,mpeghUiManager}.exe bin-audio/
     sed -i 's/^Cflags:.*/& -DMPEGHDEC_STATIC/' "$LOCALDESTDIR/lib/pkgconfig/mpeghdec.pc"
--- a/build/media-suite_compile.sh
+++ b/build/media-suite_compile.sh
@@ -1643,7 +1643,7 @@ _check=(libxavs2.a xavs2_config.h xavs2.{h,pc})
 [[ $standalone = y ]] && _check+=(bin-video/xavs2.exe)
 if [[ $bits = 32bit ]]; then
     do_removeOption --enable-libxavs2
-elif { [[ $avs2 = y ]] || { [[ $ffmpeg != no ]] && enabled libxavs2; }; } &&
+elif { [[ $avs2 != n ]] || { [[ $ffmpeg != no ]] && enabled libxavs2; }; } &&
     do_vcs "$SOURCE_REPO_XAVS2"; then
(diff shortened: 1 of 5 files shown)


ambitiuswisdom:
## Follow-up changes

- Add a reusable helper for prefixing internal symbols in static archives.
- Apply symbol prefixing to libmpeghdec only when libfdk-aac is enabled.
    - Preserve public MPEG-H APIs and special linker/compiler symbols.
    - Prevent bundled FDK symbols from colliding with libfdk-aac.

build/media-suite_compile.sh
```diff
@@ -1283,6 +1283,33 @@ if [[ $ffmpeg != no ]] && enabled liblc3 &&
...
 
+_check=(libmpeghdec.a mpeghdec.pc mpeghdec/{mpeghexport,mpeghdecoder}.h)
+[[ $standalone = y ]] && _check+=(mpeghdec/mpeghUIManager.h
+    bin-audio/{mpeghDecoder,mpeghUiManager}.exe)
+if [[ $ffmpeg != no ]] && enabled libmpeghdec &&
+    do_vcs "$SOURCE_REPO_MPEGHDEC"; then
+    do_uninstall include/mpeghdec "${_check[@]}"
+    # Avoid duplicate SHORT overloads with MinGW's LLP64 data model.(Upstream issue)
+    grep_or_sed '__x86_64__) && !defined(_WIN32)' src/libFDK/include/common_fix.h \
+        's;!defined\(_MSC_VER\) && defined\(__x86_64__\);& \&\& !defined(_WIN32);'
```
1480c1: fix was merged upstream 6 hours ago

1480c1 (changes requested):
> Configurable CPU target

This should be something the user adds to `/local64/etc/profile2.local` rather than something we ask for and set, and should only really be set by users who understand what the baseline means and why it exists.

Something like
```
#!/usr/bin/bash
CFLAGS+=" -march=native"
RUSTFLAGS+=" -C target-cpu=native"
```

> Restore libssh support

I think it looks fine, but I haven't inspected libssh's pc file yet.

> Optional 10-bit davs2

Although I am hesitant to add more forks as options after the whole zlib with chromium or rust issues, I think it should be okay considering it's adding a new feature.

ambitiuswisdom:
Revert targetPlatform option and mpeghdec patch

PR #3185, opened 2026-09-16
feat: add libastcenc

ambitiuswisdom:
ffmpeg add libastcenc support in https://github.com/FFmpeg/FFmpeg/commit/6e9100c64894728cd92aaf8ce2f1f33cfdbc675f . Therefore, also add libastcenc support in this suite.


1480c1:
Please add the option to the bat file.
I assume astcenc doesn't provide a pkg-config file for FFmpeg to pick up on?

ambitiuswisdom:
ffmpeg's build system does not pick up astcenc lib and header file by pkg-config file. Instead, it just relays on direct import and link parameter. See: https://github.com/FFmpeg/FFmpeg/blob/976cf5f9120468f1a3af12aee41a817918fac977/configure#L7197-L7208
Also, astcenc recommends integrating code to use the lib, so its cmake system doesn't generate pkg-config file. It even doesn't install static lib and header file either.

As for bat file, should I add a separate option to determine whether to enable astcenc, or just add it to `ffmpeg_options_full` list, or do both?

1480c1:
Just add it to the full options list for now. I don't think it's worth having a separate option at this time.

ambitiuswisdom:
I have synced all missing libs in bat/doc from `media-suite_compile.sh`.

